PR #583, opened 2014-08-12
Missing message when request new password 

griotteau:
- Success message not displayed when new password email is sent
- error message not displayed when store email is unknown


bibich:
It was been fixed in #730.
Thanks.

PR #685, opened 2014-09-05
Modules routes priorities

bilhackmac:
Allow modules to overwrite routes according to their positions. Greater the module position, greater route will be matched first

### DIFF
--- a/core/lib/Thelia/Model/ModuleQuery.php
+++ b/core/lib/Thelia/Model/ModuleQuery.php
@@ -12,8 +12,6 @@ use Thelia\Module\BaseModule;
  * You should add additional methods to this class to meet the
  * application requirements.  This class will only be generated as
  * long as it does not already exist in the output directory.
- *
- *
  */
 class ModuleQuery extends BaseModuleQuery
 {


lopes-vincent:
I agree with @bilhackmac.
I have the same issue (ok it's the same project).
I think this PR is good because :

1) What your are saying is if I want create a module who override routing and form of another module for example, I have to declare my router and define his priority. The method for do that is not in the doc and I have to know it's possible.

2) You said the position of module is just for payment position. Sorry but no, on another project I sort the module by priority because I have to override some form but by chance no route.
And it's logic that the routing priority have to be the same by default than the other things without declare router and change his priority.

lovenunu:
You fall back in the FIFO / LIFO problematics
I love stacks. LIFO FTW.

lunika:
@bilhackmac paid you for posting this message. I know it.

bilhackmac:
@lopes-vincent : It's not for the same project
@lunika : not at all ! I even explained undocumented router configuration but it's not as usable as you think :kissing_heart:

lopes-vincent:
@bilhackmac the same client then?
@lunika I'm not for sale! ;)
I got stuck today because I have to do a route in module that override another route from another module (more generic) at the beginning I wonder why my new route doesn't override the old one.
And after I realized this is because my modules was under the old modules so I decide to put him over it and my new route work but my listener who have also to override the generic module doesn't override it anymore.
So I think this is a real problem and this PR solve it.

lunika:
The documentation is now up-to-date http://doc.thelia.net/en/documentation/modules/routing.html

lovenunu:
So we can close the closed PR. Thank you !

bilhackmac:
You can now use [this module](https://github.com/bilhackmac/RoutesMatchInverter) to get the same behavior as this PR.

PR #1026, opened 2014-12-31
Fix the module skeleton location when creating module

bcabarrou:
When using the fancy new way of installing Thelia with componants managed by Composer, the module:create command breaks.
This is because the command assume that the Thelia core will always be in THELIA_ROOT, which is not always true anymore.

This change instead builds the skeleton directory path from the command file directory.


lunika:
Thanks.

Best coin

PR #1065, opened 2015-01-20
Added missing form field type

bibich:


core/lib/Thelia/Controller/Admin/AdminLogsController.php
```diff
@@ -41,11 +43,12 @@ public function loadLoggerAjaxAction()
         ) as $entry) {
             $entries[] = array(
                 "head" => sprintf(
-                    "[%s][%s][%s:%s]",
+                    "[%s][%s][%s:%s%s]",
```
lunika: why there is no white space before the new parameter ? Do you really want to keep it like this ?

Thanks

lunika:
Your will be merged in the version 2.2.0-alpha1 so you have to change the sql update file. Moreover don't forget to update the mysql workbench schema

core/lib/Thelia/Controller/Admin/AdminLogsController.php
bibich: Because it can be empty. So the colon is added in the argument when it 's not empty.

PR #1072, opened 2015-01-21
Add the customers/checkLogin API endpoint

bcabarrou:
This endpoint allows to check if an email and password are valid login credentials for a customer, and retrieve this customer if they are.

Some thoughs and questions on HTTP return codes:
- a success (customer exists) is a 200
- a malformed query will be a 500 for consistancy with the other API methods, but maybe it should be a 400 (Bad Request)
- an email not in use will be a 404
- an incorrect password will also be a 404 since we check if (email, password) map to a customer. Or should it be 403 ?

The method also call the getAction method to get the customer if it exists, is it OK ?

A PR to update documentation will follow.


bcabarrou:
Documentation PR at https://github.com/thelia/thelia.github.io/pull/113

lovenunu:
Unit tests :joy: Thanks !

lunika:
Thank you !

PR #1075, opened 2015-01-21
Add render function for smarty

lovenunu:
This implements a ```render``` function for Smarty.
It works like symfony twig extension.

How to use it ?
---

the render function takes a mandatory argument ```action```, that is the method of the action. It supports ```a:b:c``` notation. 
Example:
```smarty
{render action="Module:Foo:bar"}
```
Will executes ```Module\Controller\FooController::barAction``` and print the result.

__You can specify the request method__:
```smarty
{render action="Module:Foo:bar" method="PUT"}
```

__You can specify the query parameters__:
```smarty
{render action="Module:Foo:bar" query="foo=bar&baz=thelia"}
{render action="Module:Foo:bar" query=$anArray}
```

__Same for request__ _(if the method isn't specified, it will automaticly set it to POST)_:
```smarty
{render action="Module:Foo:bar" request="foo=bar&baz=thelia"}
{render action="Module:Foo:bar" request=$anArray}
```

All the other parameters will be used as the method parameters.
You can execute the following controller:
```php
<?php

namespace Module\Controller\FooController;

use Thelia\Controller\Front\BaseFrontController;
use Thelia\Core\HttpFoundation\Response;

class TestController extends BaseFrontController
{
    public function barAction($paramA, $paramB)
    {
        return new Response($paramA . $paramB);
    }
}
```

with then method:
```smarty
{render action="Module:Foo:bar" paramA="foo" paramB="bar"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Add the 'images-folder' tag into module.xml file to deploy the modules images
 - Add the 'module:list' command, that shows the modules state 
 - Update Admin Logs to add the resource ID when available.
+- Add render smarty function, that executes the controller given in the action parameter.
 
 # 2.1.2
 
--- a/local/modules/TheliaSmarty/Config/config.xml
+++ b/local/modules/TheliaSmarty/Config/config.xml
@@ -66,6 +66,13 @@
             <tag name="thelia.parser.register_plugin"/>
         </service>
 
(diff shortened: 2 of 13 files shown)


lunika:
can you rebase your branch please ?

Thanks

lovenunu:
Done !

Issue #1087, opened 2015-01-26
"Add a module to a hook" is unusable by most users.


roadster31:
Is it possible to sort hook names in alphabetical order, and also append the hook code in the label, so that one's can find easily the wanted hook directly from the name used in the template ?

Additionally, is there a way to simplify hook creation so that Service ID  and Method names will be automatically guessed or retrieved from module configuration ? Most users are unable to find these informations, as watching the source code seems the only way to get the Method Name, and reading XML the only way to get a service ID (whis is not, in the config file a "service" ID, but a "hook" ID).

Maybe some additional module configuration will help ?

roadster31:
An additional PR will be done to simplify hook creation.

PR #1089, opened 2015-01-26
Adding/updating a module hook is now working

roadster31:
This PR fixes creation and update of a module hook, and offers a better hook select menu in create and update Module Hook (#1087)


core/lib/Thelia/Form/ModuleHookCreationForm.php
```diff
@@ -19,6 +19,7 @@
 use Thelia\Core\Translation\Translator;
 use Thelia\Model\Hook;
 use Thelia\Model\HookQuery;
+use Thelia\Model\Map\HookI18nTableMap;
```
lunika: this is class is never used.

roadster31: Fixed.

PR #1096, opened 2015-01-27
Images may be zoomed before crop, to match required image size

roadster31:
This feature is enabled by setting the new Image loop `allow_zoom` parameter to 1. This parameter is off by default.


lunika:
can you update the changelog please ?

Thanks

lunika:
and after you have to rebase your branch, it's not possible to merge it right now.

Thanks

PR #1100, opened 2015-01-29
Throw assets errors when in development mode

roadster31:
When in development mode, an exception is thrown when an error occurs when processing assets, thus helping to diagnose missing files, LESS syntax errors, and the like.

In production mode, nothing changes.

A LESS syntax error

image 685

Missing asset : 

image 686


lunika:
:+1: Thanks

PR #1109, opened 2015-01-30
New module_config smarty function

roadster31:
This PR adds a new Smarty function `module_config`, to get the value of a module's configuration parameter. This parameter should be stored in the ModuleConfig table.

Usage example : 

```smarty
{module_config module="carousel" key="autostart"}
```
The useless plugin TheliaSyntax has been removed.


lunika:
Thanks :+1: 

Would be great if you prepare the documentation for this nes feature <3

PR #1119, opened 2015-02-02
Add exit after header to stop the script. Fix #538

### LINKED ISSUE #538: Exception in the install process if the admin passwords don't match
In end.php, if the admin passwords don't match, the following error occurs : 

`ErrorException: Warning: Cannot modify header information - headers already sent by (output started at /homepages/7/d221652477/htdocs/thyno/web/install/header.php:58) in /homepages/7/d221652477/htdocs/thyno/web/install/end.php line 27`

lovenunu:


lovenunu:
To complete this PR, we could add an ```ob_start();``` in header.php, before everything, and an ```ob_flush();``` in footer.php. That would prevent any "header already sent" problem, don't you think ? @lunika @roadster31

roadster31:
Yes.

lovenunu:
@lunika Finished && rebased

PR #1124, opened 2015-02-03
fix placeholder for phone and cellphone

gillesbourgeat:
Hello,

The PR is for fix placeholder phone and cellphone in tpl BO customer-edit.html


lunika:
Thanks !

PR #1127, opened 2015-02-04
Suggestion : add default behavior loop

gillesbourgeat:
Hello,

This request follows a problematic.

It is sometimes difficult to know if the behavior of a loop was changed.
If you want to be sure to have the original behavior of the loop, then it must extend and give it a new name.

With this pull request, you can quickly find the original loop behavior.

Exemple : 
{loop type="product"}
  // behavior not necessarily known
{/loop}

{loop type="product.default"}
  // default behavior
{/loop}

### DIFF
--- a/core/lib/Thelia/Config/Resources/loop.xml
+++ b/core/lib/Thelia/Config/Resources/loop.xml
@@ -5,6 +5,7 @@
         xsi:schemaLocation="http://thelia.net/schema/dic/config http://thelia.net/schema/dic/config/thelia-1.0.xsd">
 
     <loops>
+        <!-- original loop -->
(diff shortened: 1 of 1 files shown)


lunika:
This idea is really interesting, but it will probably a pain to maintain it.

* entities are duplicated.
* you can still override de default behavior (nothing prevent this).
* we don't have to create the default entry when we will create a new loop.

BTW I think it can be possible to automate all of this and prevent non genuine behavior.

Thanks for your suggestion, I really appreciate it.

gillesbourgeat:
Hello,

Indeed, I search solutions for these problems.

Thanks

PR #1133, opened 2015-02-07
htaccess new rule

gillesbourgeat:
Hello,

new rule to avoid duplicate content.
exemple rule for remove or add www in url


lunika:
Thanks

PR #1141, opened 2015-02-10
The PaymentModule log default level is now INFO instead of ERROR

roadster31:
And that's all !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Do not register previous url on XmlHttpRequest
 - Fix deploy image directory destination
 - Fix redirect response if a AuthenticationException is catched
+- The PaymentModule log default level is now INFO instead of ERROR
 
 # 2.1.1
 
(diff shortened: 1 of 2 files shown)


lunika:
It will be better to retrieve the global configuration, what do you think ?

roadster31:
No, as the log configuration is currently hardcoded.

lunika:
It is possible to retrieve this specific configuration from the database, no ?

roadster31:
Yes, maybe in the future.

lunika:
The idea is to add the ability to change dynamically the log level using the existing ```tlog_level``` entry in the config table like this : 

```
$this->log->setLevel(ConfigQuery::read('tlog_level',Tlog::INFO));
```

Can you please complete the changelog ?

Thanks

roadster31:
Done !

roadster31:
Well, func. tests failed on 2 of 3 builds, still can't restart them. Can you do it for me @lunika please ?

lunika:
Thanks

Issue #1145, opened 2015-02-10
Can't delete a Payment/Delivery module referenced by an order.


roadster31:
A payement or delivery module can't be deleted as long as it has been used in a order.

The cryptic message ``Propel\Runtime\ActiveQuery\ModelCriteria::delete is unable to delete. ` ìs displayed to the end-user when he/she's trying to delete the module.

The quick way to solve this issue is to display a meaningful error message.

However, we have to find a better solution, so that modules could be deleted even if they are currently referenced in some orders, as during the life of an e-shop, payment and delivery modules may change several times, leaving the module list stuffed with obsolete modules.
Maybe something like an `order_module table` ?

bibich:
We should add the possibility to deactivate a delivery/payment module when customers place orders, but keep it activated everywhere else. Because modules can use hooks and add information in the invoice, customer account, ...

anoziere:
First half shipped, second half still open.

The cryptic `ModelCriteria::delete is unable to delete.` you reported is no longer reachable: the guard was committed the day this issue was opened (d011808f, February 2015) and is still there on both lines (`core/lib/Thelia/Action/Module.php:218` on `2.6`, `:259` on `main`). It throws a translated message before the delete, so the Propel error never surfaces. What the message did not say is what to do instead.

That is what #3546 (`2.6`) and #3548 (`main`) add: the message now tells the administrator to deactivate the module rather than delete it, with English and French. The nine existing translations of the first sentence are untouched.

The real request stays open: a link table between order and module, so a payment or delivery module that a shop no longer uses can actually be removed. Two things found while looking at this, worth knowing before that work starts:

- The two blocking foreign keys are the only `ON DELETE RESTRICT` ones pointing at `module` (setup/thelia.sql:774-783 on `main`). The nine others cascade.
- Reinstalling a module from a zip has the same problem today. `Action\Module::install()` dispatches `MODULE_DELETE` for the previous version and only swallows the failure when the module directory is already gone (`main` core/lib/Thelia/Action/Module.php:375-381). Since the guard throws before the directory is removed, upgrading a payment module that an order references is refused, and the administrator gets a message about deletion while trying to install.

anoziere:
Instructed in depth, and **the premise of this issue turns out to be wrong in a way that matters more than the deletion itself**.

[...]

PR #1146, opened 2015-02-10
Display a meaningful message when deleting a module referenced by an order.

roadster31:
This is a TEMPORARY fix for #1145. This issue should not be closed, as we have to find a way to delete modules even if they're referenced by existing orders.


lunika:
Thank you

Issue #1088, opened 2015-01-26
Modules can't be removed from a Hook


roadster31:
In  Home / Configuration / Hooks position, it is not possible to remove a module from a Hook. When clicking the trash icon of a module, the page is reloaded, but the module is still attached to the hook.

roadster31:
After some investigation, `RegisterListenersPass::registerHook()` always create module hooks tagged with `hook.event_listener` ìf they don't exist in the database.

Thus, deleting a tagged module hook is not possible at the moment, as such hooks are immediately re-created.

I'm thinking about adding a table (such as IgnoredModuleHooks) to store the deleted module hooks. This table will be checked by `RegisterListenersPass::registerHook()` to prevent a re-creation.

Any thoughts, guys ?

lovenunu:
Add an "activate" column in the appropriate table would be easier :)

roadster31:
Which table would be "appropriate" ? If you think about ModuleHook table, we will have to refactor and maybe refactor each of the ModuleHookQuery calls.

lunika:
@roadster31 what would be the best ?

if the hook is present in your new table it is not needed to add it in ModuleHook, am I right ?

We have to create a new page for managing removed tagged module.

roadster31:
I'll go for a new table. Deleting a module from a hook will just add the (module, hook) couple to this table. Adding again the module to the hook will remove the couple from the table.

A specific page is not required.

PR #1161, opened 2015-02-14
Fixes for #1088 and #1087

roadster31:
Deleting a module from a hook will just add the (module ID, hook ID) couple to the new `ignored_module_hook` table, thus preventing `RegisterListenersPass::registerHook()` to re-create module hooks  in the database (#1088).

Adding the module to the hook will remove the couple from the table.

The module hook creation and edition will now get classname and method from the database instead of leaving the user entering them (#1087).


lunika:
This PR will be merged in branch 2.1 so you can revert the modification made in the file 2.2.0-alpha1.sql, this modifications are not needed

roadster31:
Done (and squashed)

PR #1167, opened 2015-02-15
Fixed language when previewing e-mails

roadster31:
When previewing email, the current edition language was ignored, and the en_US version was always displayed.

This PR fixes the problem.


roadster31:
Done.

PR #1169, opened 2015-02-16
Adding Pinterest to social hook

enurit:


lunika:
As you use the translator, you have to translate this string at least in french : 

* Pinterest username

PR #1173, opened 2015-02-16
Use of BaseController::createForm() instead of explicit form instantiation

roadster31:
All core form explicit instantiations (`$form = new SomeForm();`) are replaced by calls to `BaseController::createForm()`, (`$form = $this->createForm('some.form');`)

Three sets of constants are created (AdminForm, FrontForm and ApiForm), to define a constant for each of the core forms.

Minor CS fixes in some places.


lunika:
:clap:

roadster31:
:metal:

lovenunu:
:+1: :heart:

PR #1176, opened 2015-02-17
check escaping variable in templates

lunika:
some variables were not well escaped in templates


roadster31:
There are a lot of useless/unnecessary quotes,  like in `product="{$product_id}"`,  which could be written `product={$product_id}`

lunika:
I don't agree with you when you say "if escaping is not required.". You must escape a string (or whatever variable type, because you don't know the type of the variable) when you use it in the intl function because the intl function can't use the auto escaping.

For the quote, this can be true but with or without quotes you have the same results.

PR #1183, opened 2015-02-18
Add cache on the graph of the home page

gillesbourgeat:
Add cache on the graph of the home page.
Possibility to disable cache or change ttl cache, with the configuration variable admin_cache_home_stats_ttl

Default value for admin_cache_home_stats_ttl = 600

If admin_cache_home_stats_ttl == 0, then the cache is disabled.

Add btn for clear cache in graph toolbar.

Cheers,

### DIFF
--- a/core/lib/Thelia/Controller/Admin/HomeController.php
+++ b/core/lib/Thelia/Controller/Admin/HomeController.php
@@ -57,11 +57,10 @@ class HomeController extends BaseAdminController
         if ($cacheEnable) {
             $context = "_" . $this->getRequest()->query->get('month', date('m')) . "_" . $this->getRequest()->query->get('year', date('Y'));
 
-            $cacheDir = $this->getCacheDir();
             $cacheKey = self::STATS_CACHE_KEY . $context;
             $cacheExpire = intval(ConfigQuery::read("admin_cache_home_stats_ttl", '600'));
 
-            $cacheDriver = new FilesystemCache($cacheDir);
+            $cacheDriver = new FilesystemCache($this->getCacheDir());
 
             if (!$this->getRequest()->query->get('flush', "0")) {
                 $cacheContent = $cacheDriver->fetch($cacheKey);
@@ -143,7 +142,7 @@ class HomeController extends BaseAdminController
     {
         $cacheDir = $this->container->getParameter("kernel.cache_dir");
         $cacheDir = rtrim($cacheDir, '/');
-        $cacheDir .= '/' . self::STATS_CACHE_DIR . '/';
+        $cacheDir .= DIRECTORY_SEPARATOR . self::STATS_CACHE_DIR . DIRECTORY_SEPARATOR;
 
         return $cacheDir;
     }
--- a/core/lib/Thelia/Controller/Admin/HomeController.php
+++ b/core/lib/Thelia/Controller/Admin/HomeController.php
@@ -50,15 +50,14 @@ class HomeController extends BaseAdminController
             return $response;
         }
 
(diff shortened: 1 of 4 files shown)


core/lib/Thelia/Controller/Admin/HomeController.php
```diff
@@ -74,61 +86,100 @@ public function loadStatsAjaxAction()
             return $response;
         }
 
-        $data = new \stdClass();
+        $cacheEnable = intval(ConfigQuery::read("admin_cache_home_stats_enable", '1'));
```
lunika: I see you use a lot the config table in your PR and it's a good idea.

It's a good practice to create an alias when you want to retrieve the value of a specific config. Doing this you prevent typo and you can easily change the name of your variable without refactoring all your code. Moreover, each time you cast the result in int. Do this in the alias method.

++

core/lib/Thelia/Controller/Admin/HomeController.php
```diff
@@ -74,61 +86,100 @@ public function loadStatsAjaxAction()
...
+        $cacheContent = false;
 
-        /* sales */
-        $saleSeries = new \stdClass();
-        $saleSeries->color = $this->getRequest()->query->get('sales_color', '#adadad');
-        $saleSeries->data = OrderQuery::getMonthlySaleStats(
-            $this->getRequest()->query->get('month', date('m')),
-            $this->getRequest()->query->get('year', date('Y'))
-        );
+        if ($cacheEnable) {
```
lunika: variable cacheEnable is used only once, so use directly the ConfigQuery method here.

core/lib/Thelia/Controller/Admin/HomeController.php
```diff
@@ -74,61 +86,100 @@ public function loadStatsAjaxAction()
...
 
-        /* new customers */
-        $newCustomerSeries = new \stdClass();
-        $newCustomerSeries->color = $this->getRequest()->query->get('customers_color', '#f39922');
-        $newCustomerSeries->data = CustomerQuery::getMonthlyNewCustomersStats(
-            $this->getRequest()->query->get('month', date('m')),
-            $this->getRequest()->query->get('year', date('Y'))
-        );
+            $cacheKey = self::STATS_CACHE_KEY . $context;
+            $cacheExpire = intval(ConfigQuery::read("admin_cache_home_stats_ttl", '600'));
```
lunika: same here for the alias

gillesbourgeat:
Hello,

is there really a need to create a static method in config class for a variable that will  be used only here ?

lunika:
@gillesbourgeat yes it is. Tomorrow this variable can be used anywhere and if we use a string we can introduce a typo and it's hard to detect this type of error.

core/lib/Thelia/Controller/Admin/HomeController.php
```diff
@@ -38,61 +50,99 @@ public function loadStatsAjaxAction()
...
-            $newCustomerSeries,
-            $orderSeries,
-            $firstOrderSeries,
-            $cancelledOrderSeries,
-        );
-
-        $json = json_encode($data);
-
-        return $this->jsonResponse($json);
+        $cacheExpire = intval(ConfigQuery::getAdminCacheHomeStatsTTL());
```
lunika: with this static method you can control what is returned, please put your intval in the method ```getAdminCacheHomeStatsTTL()```

lunika: Better, you can cast it in boolean if you want

gillesbourgeat: Not possible to boolean, because it can be a time value.

lunika:
:+1:

PR #1184, opened 2015-02-19
Added the failsafe parameter to assets processing

roadster31:
When true, the `failsafe` parameter prevent the re-throw of exceptions when an asset is not found, even in development mode.

Example usage of this parameter: the Markdown plugin optionnal language file : 

```smarty
{javascripts file="assets/locale/bootstrap-markdown.{lang attr="code"}.js" source="Markdown" failsafe=true}
    <script src="{$asset_url}"></script>
{/javascripts}
```


lunika:
Can you complete the changelog ?

Thanks

PR #1186, opened 2015-02-19
Assets filters management improvement

roadster31:
To provide a flexible and efficient way to manage assets filters, the asset filters are now declared in the container configuration, using the `thelia.asset.filter` tag.

### DIFF
--- a/core/lib/Thelia/Core/Template/Assets/AsseticAssetManager.php
+++ b/core/lib/Thelia/Core/Template/Assets/AsseticAssetManager.php
@@ -217,9 +217,10 @@ class AsseticAssetManager implements AssetManagerInterface
                 foreach ($this->assetFilters as $filterIdentifier => $filterInstance) {
                     if ($filterIdentifier == $filter_name) {
                         $filterManager->set($filterIdentifier, $filterInstance);
+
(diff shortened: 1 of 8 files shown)


core/lib/Thelia/Core/Template/Assets/AsseticAssetManager.php
```diff
@@ -212,39 +214,18 @@ protected function decodeAsseticFilters(FilterManager $filterManager, $filters)
...
-
-                    case 'cssimport':
-                        $filterManager->set('cssimport', new Filter\CssImportFilter());
-                        break;
+                foreach ($this->assetFilters as $filterIdentifier => $filterInstance) {
+                    if ($filterIdentifier == $filter_name) {
+                        $filterManager->set($filterIdentifier, $filterInstance);
+                    }
+                    // No, goto is not evil.
+                    goto filterFound;
```
lunika: I don't know if it's evil or not but are you sure to use it here ? Why don't in the if statement just before ?

Here I will always go to filterfound, don't you think ?

roadster31: Nice catch.

PR #1189, opened 2015-02-21
Display image not found in template image upload list

gillesbourgeat:
Display image not found in template image upload list, if there is an error during the processiong generation.

Currently, if the image is not found on the file system, it is not present in the list.

Cheers,

### DIFF
--- a/templates/backOffice/default/includes/image-upload-list-ajax.html
+++ b/templates/backOffice/default/includes/image-upload-list-ajax.html
@@ -13,7 +13,7 @@ Parameters:
 
 {ifloop rel="image"}
     <ul id="js-sort-image" class="row list-unstyled">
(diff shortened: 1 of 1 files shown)


templates/backOffice/default/includes/image-upload-list-ajax.html
```diff
@@ -13,7 +13,7 @@
 
 {ifloop rel="image"}
     <ul id="js-sort-image" class="row list-unstyled">
-        {loop type="image" name="image" source=$imageType order="manual" source_id=$parentId width="200" height="100" resize_mode="borders" visible="*"}
+        {loop type="image" name="image" backend_context="on" ignore_processing_errors="on" source=$imageType order="manual" source_id=$parentId width="200" height="100" resize_mode="borders" visible="*"}
```
lunika: setting backend_context to on is enough I think.

PR #1200, opened 2015-02-24
Default product template in categories

roadster31:
This PR add a  new feature: a default product template could be defined in categories. 

Products created in this category will get this default product template. If no default product template is defined in a given category, it will be searched in parent categories.


core/lib/Thelia/Controller/Admin/CategoryController.php
```diff
@@ -24,6 +24,7 @@
 use Thelia\Core\Security\AccessManager;
 use Thelia\Core\Security\Resource\AdminResources;
 use Thelia\Form\Definition\AdminForm;
+use Thelia\Model\Category;
```
lunika: This class is never used

roadster31: It is. Line 119 (same old debate, isn't it ?)

tests/phpunit/Thelia/Tests/Action/CategoryTest.php
```diff
@@ -16,9 +16,12 @@
 use Thelia\Core\Event\Category\CategoryDeleteEvent;
 use Thelia\Core\Event\Category\CategoryToggleVisibilityEvent;
 use Thelia\Core\Event\Category\CategoryUpdateEvent;
+use Thelia\Core\Event\Template\TemplateDeleteEvent;
 use Thelia\Model\Category as CategoryModel;
 use Thelia\Core\Event\Category\CategoryCreateEvent;
 use Thelia\Model\CategoryQuery;
+use Thelia\Model\Template;
+use Thelia\Model\TemplateQuery;
```
lunika: this class is never used

lunika:
please update the file ```setup/update/sql/2.2.0-alpha.sql``` with the new modification.

Thanks

PR #1201, opened 2015-02-25
Automatic generation of form unique name in BaseFom

roadster31:
This PR improve BaseForm by generating the form unique name from the full class name.

The getName() method is now concrete, and may be overridden if required.


lovenunu:
:+1:

PR #1209, opened 2015-02-28
Improved countries modification

roadster31:
This PR improves countries management, and remove the obsolete "Area" (e.g. Shipping Zone) field.

The form field renderer does no longer escape the 'help' field, thus allowing to put links and other code in help text.


local/modules/Front/Config/front.xml
```diff
@@ -222,7 +222,7 @@
     <!-- end contact management -->
 
     <!-- newsletter management -->
-    <route id="newsletter.process" path="/newsletter" methods="post">
+    <route id="newsletter.process" path="/newsletter">
```
lunika: why did you remove the methods ?

roadster31: I don't remember.

PR #1221, opened 2015-03-03
Fix rounding price 

Yochima:
Fix rounding price during order creation AND when we want to get totalOrderAmount

This allow to stop having different prices between Front and Back


lunika:
I have to integrate it in branch 2.1 ?

Thanks

Yochima:
Yes please :)

PR #1223, opened 2015-03-03
Add hook and fix hook templates.table-row

gillesbourgeat:
Hello,

Add new hook : 
- profile.table-row
- profile.table-header
- import.table-row
- import.table-header
- export.table-row
- export.table-header

Add parameter template_id for hook templates.table-row in backoffice -> templates.html

cheers,


lovenunu:
You don't need to use the ```location``` parameter when you write a new hook. It was used for backward compatibility of 2.0 modules ```AdminIncludes```.

but anyway, good job ;)

templates/backOffice/default/profiles.html
```diff
@@ -75,6 +76,7 @@
                                         {/if}
                                     </td>
                                     <td>{$DESCRIPTION}</td>
+                                    {hook name="profile.table-row" location="profile_table_row" sale_id=$ID}
```
bibich: It should be `profile_id` ?

gillesbourgeat:
@bibich :+1: 
@lovenunu :+1: 

This pull request will come back.

Terminator

lovenunu:
Why did you close your PR ? O.o

gillesbourgeat:
The better to reopen later.

PR #1229, opened 2015-03-05
Add thelia logger as a service

lovenunu:


roadster31:
:+1:

lovenunu:
I think I'll have to write some doc to force developpers to use it instead of the singleton

roadster31:
We'll have to deprecate TLog.

And also define a TheliaLoggerInterface.

lovenunu:
We don't need any interface. Tlog currently implements ```Psr\Log\LoggerInterface```, I think we should base our work on that one to integrate Monolog. We could translate destinations into monolog handlers, but the two loggers are so different ... I think it'll be fun to work on that :)

roadster31:
We do, as `TLog::addXXXX() methods are widely used in the code, and are not part of ``Psr\Log\LoggerInterface`

roadster31:
What about the interface ?

lovenunu:
Well, I think that rather writing our own interface, we should deprecate ```addXXX``` methods and remove them when we integrate another logger. What do you think of this @lunika and @bibich ?

roadster31:
Ir's fine for me.

PR #1232, opened 2015-03-05
Hook and modules UI improvements

roadster31:
This PR adds a pretty nifty information dialog to modules. Data are found in the DB and the module descriptor. 

The "hook positions" term, which is not really meaningful, was replaced by "modules attachments".

In the module attachments page, the module code is added along the module title.

image 087

Last but not least, in the order list view, the status viewed is no longer lost when changing the status of an order.


core/lib/Thelia/Controller/Admin/ModuleController.php
```diff
@@ -362,4 +364,40 @@ public function installAction()
...
+                    [ '%id' => $module_id]
+                );
+            }
+
+        } else {
+            $title = $this->getTranslator()->trans('Error occured.');
+            $content = $this->getTranslator()->trans('Module ID "%id" was not found.', [ '%id' => $module_id]);
+        }
+
+        return new JsonResponse([ 'title' => $title, 'body' => $content]);
```
lunika: if the module does not exists, the http status code must be 404

core/lib/Thelia/Controller/Admin/ModuleController.php
```diff
@@ -362,4 +364,44 @@ public function installAction()
...
+                $status = 404;
+
+                $content = $this->getTranslator()->trans(
+                    'Failed to load descriptor (module.xml) for module ID "%id".',
+                    [ '%id' => $module_id]
+                );
+            }
+
+        } else {
+            $title = $this->getTranslator()->trans('Error occured.');
```
lunika: the 404 error is here and not in the previous. In the previous else it's more a 500 error.

PR #1238, opened 2015-03-06
WIP Method fromApi, fromAdmin, fromFront in request service

gillesbourgeat:
Hello,

This PR is a proposition for detect if the request comes from the api, admin or front
I have not found a method to do in Thelia.

The method name and content is not final.
All propositions are always welcome.

Cheers,

### DIFF
--- a/core/lib/Thelia/Core/HttpFoundation/Request.php
+++ b/core/lib/Thelia/Core/HttpFoundation/Request.php
@@ -13,6 +13,9 @@
 namespace Thelia\Core\HttpFoundation;
 
 use Symfony\Component\HttpFoundation\Request as BaseRequest;
+use Thelia\Controller\Admin\BaseAdminController;
+use Thelia\Controller\Api\BaseApiController;
+use Thelia\Controller\Front\BaseFrontController;
 use Thelia\Model\ConfigQuery;
 
 /**
@@ -25,6 +28,9 @@ use Thelia\Model\ConfigQuery;
 class Request extends BaseRequest
 {
     private $resolvedPathInfo;
+
+    /** @var string */
+    protected $controllerType = null;
 
     /**
      * Filter PathInfo to allow slash ending uri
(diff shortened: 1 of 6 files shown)


lovenunu:
Well ... That's the same problem as find if the request is for the back office.
You can write the route ```/admin/foo``` and use a ```BaseFrontController```.
I don't think this is the solution.

lunika:
this routes are not part of the API : 
```
    <route id="admin.configuration.api.downloadSecure" path="/admin/configuration/api/secure_key/{api_id}">
        <default key="_controller">Thelia\Controller\Admin\ApiController::downloadAction</default>
    </route>

    <route id="admin.configuration.api.delete" path="/admin/configuration/api/delete">
        <default key="_controller">Thelia\Controller\Admin\ApiController::deleteAction</default>
    </route>

    <route id="admin.configuration.api.update" path="/admin/configuration/api/update/{api_id}" methods="get">
        <default key="_controller">Thelia\Controller\Admin\ApiController::updateAction</default>
    </route>

    <route id="admin.configuration.api.update.process" path="/admin/configuration/api/update/{api_id}" methods="post">
        <default key="_controller">Thelia\Controller\Admin\ApiController::processUpdateAction</default>
    </route>
```

But your method will return true.

lovenunu:
s/>=/===/g

core/lib/Thelia/Core/HttpFoundation/Request.php
```diff
@@ -87,4 +94,51 @@ public function toString($withContent = true)
...
+    }
+
+    /**
+     * Detect if the request comes from the api
+     *
+     * @return bool
+     */
+    public function isApi()
+    {
+        if ($this->controllerType === self::CONTROLLER_API) {
```
lunika: you can directly put : 

```
return $this->controllerType === self::CONTROLLER_API;
```

It's the exact same thing.

gillesbourgeat:
@lunika indeed, rebase ok.

core/lib/Thelia/Core/HttpFoundation/Request.php
gillesbourgeat: @lunika indeed, rebase ok.

lunika:
:+1:

PR #1243, opened 2015-03-09
OrderEvent dispatching fix

roadster31:
This PR fix a problem with the OrderEvent. 

This event is dispatched several times by the Order action. As this is the same instance that is dispatched, if someone call stopPropagation() during the event processing, all further dispatching of the same event instance will have no effect.

In this PR, the vent is cloned before beeing dispatched again.


roadster31:
Wrong branch.

PR #1244, opened 2015-03-09
Order event should be cloned if dispatched several times

roadster31:
This PR fix a problem with the OrderEvent.

This event is dispatched several times by the Order action. As this is the same instance that is dispatched, if someone call stopPropagation() during the event processing, all further dispatching of the same event instance will have no effect.

In this PR, the vent is cloned before beeing dispatched again.


lunika:
If the first event has the stopPropagation set to true, when you clone this object yous till have the stopPropagation set to true.

Did you missed to implement the __clone() method ?

roadster31:
This method is called by the dispatcher, thus the event stopPropagation attribute is false when entering the method.

PR #1247, opened 2015-03-09
Fix for empty string translation issue

roadster31:
This PR fixes an issue with translations of empty strings, such as `{intl l=''}`, which is not properly matched by the regular expression.


lunika:
I have to merge it in branch 2.1 ?

roadster31:
Up to you.

PR #1265, opened 2015-03-12
Allow stacked current form

lovenunu:
Add stacked current form into parser context.
It allows to have nested forms while using the new way to write forms.


roadster31:
:top:

Issue #1270, opened 2015-03-13
coupons working even when disabled


Vyndi:
Thelia version : 2.1.1

If the expiration date has not expired, the coupon will work on products even if the coupon has been disabled.

lunika:
woot this is annoying

lunika:
The fix is merged in branch 2.0, 2.1 and master and will be present in the next releases.

Thanks.

PR #1272, opened 2015-03-13
test if coupon is enabled before consuming it

lunika:
as noticed in #1270 a coupon can be used event if it's disabled.

### DIFF
--- a/core/lib/Thelia/Exception/CouponDisabledException.php
+++ b/core/lib/Thelia/Exception/CouponDisabledException.php
@@ -1,0 +1,34 @@
+<?php
+/*************************************************************************************/
+/*      This file is part of the Thelia package.                                     */
+/*                                                                                   */
+/*      Copyright (c) OpenStudio                                                     */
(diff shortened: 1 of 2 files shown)


lunika:
The translation are not made because I want to merge this PR in all branches.

core/lib/Thelia/Exception/CouponDisabledException.php
```diff
@@ -0,0 +1,34 @@
...
+use Thelia\Core\Translation\Translator;
+use Thelia\Log\Tlog;
+
+/**
+ * Class CouponDisabledException
+ * @package Thelia\Exception
+ * @author Manuel Raynaud <manu@thelia.net>
+ */
+class CouponDisabledException extends \RuntimeException
+{
```
roadster31: InactiveCouponException seems a more suitable name for this exception.

lunika: Thanks, I will change its name.

PR #1292, opened 2015-03-19
remove confirm_email in customer create form if in admin context

lunika:
I don't know if it's the more sexy way to do this but I need to remove the email_confirm field in the CustomerCreateForm if we are in admin context

### DIFF
--- a/core/lib/Thelia/Core/EventListener/FormListener.php
+++ b/core/lib/Thelia/Core/EventListener/FormListener.php
@@ -1,0 +1,89 @@
+<?php
+/*************************************************************************************/
+/*      This file is part of the Thelia package.                                     */
+/*                                                                                   */
+/*      Copyright (c) OpenStudio                                                     */
+/*      email : dev@thelia.net                                                       */
+/*      web : http://www.thelia.net                                                  */
+/*                                                                                   */
+/*      For the full copyright and license information, please view the LICENSE.txt  */
+/*      file that was distributed with this source code.                             */
+/*************************************************************************************/
+
+namespace Thelia\Core\EventListener;
+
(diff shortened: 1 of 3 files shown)


core/lib/Thelia/Core/EventListener/FormListener.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * @var ControllerResolver
+     */
+    protected $resolver;
+
+    /**
+     * @param Request $request
+     * @param ControllerResolver $resolver
+     */
+    public function __construct(Request $request, ControllerResolver $resolver)
```
lovenunu: Why don't you use ```Symfony\Component\HttpKernel\Controller\ControllerResolverInterface``` for type checking ?

core/lib/Thelia/Core/EventListener/FormListener.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $this->request = $request;
+        $this->resolver = $resolver;
+    }
+
+    public function removeEmailConfirmation(TheliaFormEvent $event)
+    {
+        $controller = $this->resolver->getController($this->request);
+
+        if ($controller[0] instanceof CustomerController) {
```
lovenunu: Why do you check this controller ? Another one can use the form in a front context. :)

lunika: For the front context you can set the value of ```customer_confirm_email``` to 0 or 1 if you want to verify the email or not

lunika:
An other implementation, more easy, could be to check if the parameter ```customer_confirm_email``` is activate or not.

lunika:
#1293 is easier and cleaner :-)

lovenunu:
:+1:

PR #1296, opened 2015-03-20
Deleting a product template is now cascaded to feature_template and

roadster31:
Deleting a product template is now cascaded to feature_template and attribute_template tables.


lunika:
You have to follow this new rules when you want to change an update script : http://doc.thelia.net/en/documentation/contribute.html#sql-scripts-modification

lunika:
You have to generate the sql file once you update the tpl file. It is explained in the documentation.

PR #1299, opened 2015-03-22
Update method jsonResponse

gillesbourgeat:
Hello,

- Add the ability to change the message of the http status for json response.
- Add auto json_encode if the parameter is array or object.

Cheers,


lunika:
I would like to deprecate this method. There is no reason to use it, the ```Thelia\Core\HttpFoundation\JsonResponse``` object is more appropriate. This helper in fact doesn't really help you.

For the message, once you have the returned object, you can set your message, I see no interest to change it in the method.

gillesbourgeat:
Ok, It was just a suggestion.

PR #1301, opened 2015-03-23
Added a fallback on default email template of a module when building mes...

bibich:
...sage

When we send mails from a module and don't use the `default` email template, if the files (html/txt) are not found in the template then the template files located in the module (under `templates/email/default/' directory) are used as a fallback.


lovenunu:
i_approve

lovenunu:
If you could add a line into the changelog. Well ... That'd be great.

PR #1303, opened 2015-03-23
Added capability to use translator in module functions preActivation and...

bibich:
The goal of this PR is to make possible to use translator in the `preActivation` and `postActivation`.
As the module is not activated, the translation catalogs of the module are not yet loaded.
So we have to add resources ~~by hand~~ manually


lunika:
What is "by hand" ?

lovenunu:
It will be merged if you wrote the code **by hand**

lunika:
Can you rebase your branch please ?

PR #1304, opened 2015-03-23
Virtual product delivery

bibich:
This PR just separates the mail sending in 2 steps. When an order goes into paid status, a new event is dispatched. This new event is listened by the `VirtualProductDelivery` module and an email with the list of download files is sent. So a module can easily change this behavior (different mails or no mails) without breaking the change status event.

Also, now, the `VirtualProductDelivery` module uses email templates.

By default, the mail will be sent only if there is at least one document to download. See #1308


lunika:
Can you complete the readme ?

Thanks

PR #1308, opened 2015-03-23
Virtual products mail sent only if there's something to download

roadster31:
When virtual products do not have any related downloadable documents, the VirtualProductDelivery module sends the mail anyway.

This PR fixes the problem, and checks the number of documents before sending the mail.


bibich:
I've integrated your changes in this PR #1304

lunika:
So can I close it ?

bibich:
Yes you can.

PR #1319, opened 2015-03-26
New thelia.template_helper service to replace TemplateHelper singleton

roadster31:
To allow modules to change the way current templates are set, this PR introduces the `thelia.template_helper` service, which offer the same services as the `TemplateHelper` singleton, in a more dynamic way. 
A TemplateHelperInterface is introduced as well, to define `thelia.template_helper` interface. The default implementation of this interface is the TheliaTemplateHelper class.

TemplateHelper singleton is now deprecated, and is just a (more or less) static facade for the `thelia.template_helper` service, to provide backward compatibility.

Some translation specific features that were previously located in TemplateHelper have been moved to a specific Translation action, triggered by two new events.

### DIFF
--- a/core/lib/Thelia/Action/Translation.php
+++ b/core/lib/Thelia/Action/Translation.php
@@ -22,7 +22,7 @@ use Thelia\Log\Tlog;
 /**
  * Class Translation
  * @package Thelia\Action
- * @author  Franck Allimant <franck@cqfdev.fr>
+ * @author Manuel Raynaud <manu@thelia.net>
  */
 class Translation extends BaseAction implements EventSubscriberInterface
 {
--- a/core/lib/Thelia/Action/Translation.php
+++ b/core/lib/Thelia/Action/Translation.php
@@ -20,9 +20,9 @@ use Thelia\Core\Translation\Translator;
 use Thelia\Log\Tlog;
 
 /**
- * Class Area
+ * Class Translation
  * @package Thelia\Action
- * @author Manuel Raynaud <manu@thelia.net>
+ * @author  Franck Allimant <franck@cqfdev.fr>
  */
 class Translation extends BaseAction implements EventSubscriberInterface
 {
@@ -60,16 +60,16 @@ class Translation extends BaseAction implements EventSubscriberInterface
      */
     protected function walkDir($directory, $walkMode, $currentLocale, $domain, &$strings)
     {
-        $num_texts = 0;
+        $numTexts = 0;
 
         if ($walkMode == TranslationEvent::WALK_MODE_PHP) {
             $prefix = '\-\>[\s]*trans[\s]*\([\s]*';
 
-            $allowed_exts = array('php');
+            $allowedExts = array('php');
         } elseif ($walkMode == TranslationEvent::WALK_MODE_TEMPLATE) {
             $prefix = '\{intl(?:.*?)l=[\s]*';
 
-            $allowed_exts = array('html', 'tpl', 'xml', 'txt');
+            $allowedExts = array('html', 'tpl', 'xml', 'txt');
         } else {
             throw new \InvalidArgumentException(
                 Translator::getInstance()->trans(
@@ -89,7 +89,7 @@ class Translation extends BaseAction implements EventSubscriberInterface
                 }
 
                 if ($fileInfo->isDir()) {
-                    $num_texts += $this->walkDir(
+                    $numTexts += $this->walkDir(
                         $fileInfo->getPathName(),
                         $walkMode,
                         $currentLocale,
@@ -101,7 +101,7 @@ class Translation extends BaseAction implements EventSubscriberInterface
                 if ($fileInfo->isFile()) {
                     $ext = $fileInfo->getExtension();
 
-                    if (in_array($ext, $allowed_exts)) {
+                    if (in_array($ext, $allowedExts)) {
                         if ($content = file_get_contents($fileInfo->getPathName())) {
                             $short_path = $this->normalizePath($fileInfo->getPathName());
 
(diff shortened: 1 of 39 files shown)


core/lib/Thelia/Core/Template/TemplateHelper.php
```diff
@@ -12,321 +12,94 @@
...
     }
 
+    /**
+     * @deprecated use TheliaTemplateHelper service (thelia.template_helper) instead.
+     */
     public static function getInstance()
     {
         if (self::$instance == null) {
-            self::$instance = new TemplateHelper();
+            throw new \LogicException("This class should be initialized before getInstance() call.");
```
lunika: If you don't want BC break, why did you throw this exception ? I think you can just tag this method as deprecated.

If you want you can trigger a notice error like here : https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Controller/BaseController.php#L565

lunika:
Nice work.

You don't want, but finally you introduced a BC break. You change the behaviour of the singleton TemplateHelper. Can you find an other way to refactor this singleton ?

core/lib/Thelia/Core/Template/TemplateHelper.php
roadster31: The singleton is supposed to be initialized by TheliaTemplateHelper when it is instanciated at bootstrap. 

Thus this exception is not thrown in normal bootstrap process. I've coded it "just in case".

lunika: And you forget to declare the property ```templateHelper``` in the class

core/lib/Thelia/Action/Translation.php
```diff
@@ -0,0 +1,237 @@
...
+     * @return number the total number of translatable texts
+     */
+    protected function walkDir($directory, $walkMode, $currentLocale, $domain, &$strings)
+    {
+        $num_texts = 0;
+
+        if ($walkMode == TranslationEvent::WALK_MODE_PHP) {
+            $prefix = '\-\>[\s]*trans[\s]*\([\s]*';
+
+            $allowed_exts = array('php');
```
lunika: Variables must be in camelCase

core/lib/Thelia/Action/Translation.php
```diff
@@ -0,0 +1,237 @@
...
+                        $strings
+                    );
+                }
+
+                if ($fileInfo->isFile()) {
+                    $ext = $fileInfo->getExtension();
+
+                    if (in_array($ext, $allowed_exts)) {
+                        if ($content = file_get_contents($fileInfo->getPathName())) {
+                            $short_path = $this->normalizePath($fileInfo->getPathName());
```
lunika: Variables must be in camelCase

core/lib/Thelia/Action/Translation.php
```diff
@@ -0,0 +1,237 @@
...
+
+                                foreach ($matches[2] as $match) {
+                                    $hash = md5($match);
+
+                                    if (isset($strings[$hash])) {
+                                        if (! in_array($short_path, $strings[$hash]['files'])) {
+                                            $strings[$hash]['files'][] = $short_path;
+                                        }
+                                    } else {
+                                        $num_texts++;
```
lunika: Variables must be in camelCase

core/lib/Thelia/Action/Translation.php
```diff
@@ -0,0 +1,237 @@
...
+     * @param  string $walkMode type of file scanning: WALK_MODE_PHP or WALK_MODE_TEMPLATE
+     * @param  string $currentLocale the current locale
+     * @param  string $domain the translation domain (fontoffice, backoffice, module, etc...)
+     * @param  array $strings the list of strings
+     * @throws \InvalidArgumentException if $walkMode contains an invalid value
+     * @return number the total number of translatable texts
+     */
+    protected function walkDir($directory, $walkMode, $currentLocale, $domain, &$strings)
+    {
+        $num_texts = 0;
```
lunika: Variables must be in camelCase

lunika:
thelia.sql file is modified but I don't see the modifications in the schema.xml file.

core/lib/Thelia/Action/Translation.php
roadster31: Should we refactor all the old code we found in each PR ?

lunika: if we can, yes !

roadster31: OK, I'll have a closer look on future PR's.

roadster31:
> thelia.sql file is modified but I don't see the modifications in the schema.xml file.

The schema.xml file in master branch is not in sync with thelia.sql : 

```
<foreign-key foreignTable="attribute" name="fk_attribute_template_id" onDelete="CASCADE" onUpdate="RESTRICT">
<reference foreign="id" local="attribute_id" />
</foreign-key>
<foreign-key foreignTable="template" name="fk_attribute_template" onDelete="CASCADE" onUpdate="RESTRICT">
<reference foreign="id" local="template_id" />
</foreign-key>
```
```
CONSTRAINT `fk_attribute_template_id`
FOREIGN KEY (`attribute_id`)
REFERENCES `attribute` (`id`)
ON UPDATE RESTRICT
ON DELETE CASCADE,
CONSTRAINT `fk_attribute_template`
FOREIGN KEY (`template_id`)
REFERENCES `template` (`id`)
```

The PR fixes the problem.

lunika:
Ok, but it's not possible to merge the PR, some conflicts are still here.

roadster31:
I'm wotking on it.

lunika:
Thanks

roadster31:
Travis build failed because of "Failed asserting that '53.7567' matches expected 53.756666666666675."

Could you please restart the bulid ?

lovenunu:
There are 1.00000000000000238345487 ways to round a number with PHP.

roadster31:
To round a float.

PR #1328, opened 2015-03-30
Imagine library updated to 0.6.2

roadster31:
- Imagine library is updated to 0.6.2, which provides a better support for transparency.
- Default border colour of images resized with resize_mode="border" is now transparent instead of opaque white.


lovenunu:
As the class ```Imagine\Image\Color``` dispear, its potentially BC break for the modules that uses this dependency.

roadster31:
This is the naked truth.

lovenunu:
Well ... Don't care, upgrade is a good thing :D

roadster31:
Well... If we have to write a support class for BC, where should we put it ?

lovenunu:
Nowhere, the ```Imagine``` NS doesn't belong to us.

lunika:
Can you rebase your PR ?

Thanks

roadster31:
OK.

Could you please try to merge PR in chronological order ?

PR #1332, opened 2015-04-01
Implement nyancat printer for phpunit

lovenunu:
This PR will force developpers to run tests before submitting anything here, as having a nyancat printer for unit test forces you to write more tests *by hand* to make it last longer !

phpunit__


lunika:
Thanks for your contribution @lovenunu

PR #1338, opened 2015-04-07
require phpunit 4.6

lunika:


lovenunu:
Why don't we just require ```~4.1```, we don't really use the new phpunit features ?

lunika:
I just test this version for php 7.  

And I just follow the phpunit documentation

PR #1364, opened 2015-04-13
Add command to set modules positions

bilhackmac:


bilhackmac:
OK I'm an ass… it's working now !

PR #1368, opened 2015-04-14
Translation improvement

bibich:
This PR declares a new translation domain **global** that will be used before any other domain. The goal is to redefine string per website and not to change original translations. 

For now, the translations made in a website replaced the original translations. So when an update was done (modules, templates, thelia) all translations were overridden by the new ones.

The new translations are now saved in the `local/I18n` directory. To keep the possibility to have different translations for the same *key*, I have implemented a simple system.  
When you translate a module in the back office for `ExampleModule` for example you have the choice to translate a key globally or specifically for this module. `globally` means that if the key is also present in an other module it will be also translated. To allow this behavior, the key of the translation contains the actual translation domain if it's a translation specific to a module. The translation file will look like this. `local/I18n/fr_FR.php` : 

```php
<?php

return array(
    'Cart' => 'Caddie',
    '[hookcart.fo.default] Cart' => 'Charette',
    '[hookcart.fo.default] Remove' => 'Enlever',
    ...
);

```

So the string **Cart** will be replaced everywhere by **Caddie**, But in the front default template of the module HookCart **Cart** will be replaced by **Charette**

This is a work in progress, and I have to improve the UI of the translation page.


lunika:
The syntax is a little bit strange but why not. What do yout think @roadster31 @lovenunu @gillesbourgeat ?

lovenunu:
Why just don't use array keys ?
return [
    "global" => [ ... ]
    "hookcart.fo.default" => [ ... ]
];

roadster31:
@lovenunu :+1:

gillesbourgeat:
@lovenunu :+1: it also allows to detect entries to be deleted  for a specific translation domain.

Otherwise, it's a good job and solves a big problem for updates and external modules in a project.

lunika:
I tested it yesterday and I'm convinced to integrate this feature.

What is missing to finish this PR ?

bibich:
I think this PR is finished.

PR #1372, opened 2015-04-15
A Template Context to ease form error processing

roadster31:
This PR introduces a template context, stored in the parser context, so that a controller may get `the last rendered template, and its parameters.

This is very useful to process form errors, where the current template should be redisplayed without redirecting.

```
return $this->render(
    $this->getParserContext()->getCurrentTemplateContext()->getName(),
    $this->getParserContext()->getCurrentTemplateContext()->getParameters()
);
```

### DIFF
--- a/core/lib/Thelia/Core/Template/ParserContext.php
+++ b/core/lib/Thelia/Core/Template/ParserContext.php
@@ -12,6 +12,7 @@
 
 namespace Thelia\Core\Template;
 
+use Thelia\Core\HttpFoundation\Session\Session;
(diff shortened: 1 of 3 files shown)


lunika:
:+1:

roadster31:
This PR is not working, as Hooks also uses the render() method to get template fragments.

lunika:
I have to revert it or you want to complete it ?

roadster31:
I'd like to complete it.

roadster31:
Please revert it.

lunika:
Done

PR #1382, opened 2015-04-19
A better way to handle user input errors in Thelia forms

roadster31:
This PR offers a better way to handle user input errors in Thelia forms. 

The form status is persisted across requests (during a time limited to 60 seconds, to prevent persistence of obsolete data), so that a form which contains errors could be redisplayed even if the user browser has been redirected.


lovenunu:
wooo :+1:

lunika:
Thanks

PR #1392, opened 2015-04-22
add envrionment depends on Tinymce Module

Soldras:
Allow Tinymce to save and get images or documents from different environment.

By default all files are save in web/media/upload folder, now you can set THELIA_ENV var in your vhost to save files in web/media/THELIA_ENV/upload.

### DIFF
--- a/local/modules/Tinymce/Resources/js/tinymce/filemanager/config/config.php
+++ b/local/modules/Tinymce/Resources/js/tinymce/filemanager/config/config.php
@@ -2,21 +2,48 @@
 use Thelia\Core\HttpFoundation\Request;
 use Thelia\Core\Thelia;
 use Thelia\Model\ConfigQuery;
+use Symfony\Component\Filesystem\Filesystem;
 
-$env = 'prod';
+function generateFolder($env)
+{
+    $webMediaPath = THELIA_WEB_DIR.'media';
+    $webMediaEnvPath = null;
+    if ($env !== "prod") {
+        //Remove separtion between dev and prod in particular environment
+        $env = str_replace('_dev', '', $env);
+        $webMediaEnvPath = $webMediaPath.DS.$env;
+    }
 
-if (file_exists(__DIR__ . '/../../../../../../../../core/vendor/autoload.php')) {
+    $fileSystem = new Filesystem();
+
+    // Create the media directory in the web root , if required
+    if (null !== $webMediaEnvPath) {
+        if (false === $fileSystem->exists($webMediaEnvPath)) {
+            $fileSystem->mkdir($webMediaEnvPath.DS.'upload');
+            $fileSystem->mkdir($webMediaEnvPath.DS.'thumbs');
+        }
+    } else {
+        if (false === $fileSystem->exists($webMediaPath)) {
+            $fileSystem->mkdir($webMediaPath.DS.'upload');
+            $fileSystem->mkdir($webMediaPath.DS.'thumbs');
+        }
+    }
+}
+
+$env = getenv('THELIA_ENV') ?: 'prod';
+
+if (file_exists(__DIR__.'/../../../../../../../../core/vendor/autoload.php')) {
     // Symlinked with std install
-    require_once __DIR__ . '/../../../../../../../../core/vendor/autoload.php';
-} elseif (file_exists(__DIR__ . '/../../../../core/vendor/autoload.php')) {
+    require_once __DIR__.'/../../../../../../../../core/vendor/autoload.php';
+} elseif (file_exists(__DIR__.'/../../../../core/vendor/autoload.php')) {
     // Hard copy with std install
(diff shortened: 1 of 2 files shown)


local/modules/Tinymce/Tinymce.php
```diff
@@ -41,6 +43,16 @@ public function postActivation(ConnectionInterface $con = null)
     {
         $fileSystem = new Filesystem();
 
+        //Check for environment
+        if ($env = $this->getContainer()->getParameter('kernel.environment')) {       
+            //Check for backward compatibility
+            if(strcmp($env,"prod") !== 0){
```
lovenunu: As we benchmarked together, This is clearly not as efficient as ```$env === "prod"``` :-1:

local/modules/Tinymce/Tinymce.php
```diff
@@ -51,11 +63,18 @@ public function postActivation(ConnectionInterface $con = null)
...
             }
         }
-
-        // Create the media directory in the web root, if required
-        if (false === $fileSystem->exists($this->webMediaPath)) {
-            $fileSystem->mkdir($this->webMediaPath . DS . 'upload');
-            $fileSystem->mkdir($this->webMediaPath . DS . 'thumbs');
+        
+        // Create the media directory in the web root , if required
+        if(null != $this->webMediaEnvPath){
```
lovenunu: !==

local/modules/Tinymce/Tinymce.php
```diff
@@ -51,11 +63,18 @@ public function postActivation(ConnectionInterface $con = null)
...
+            if(false==  $fileSystem->exists($this->webMediaEnvPath)){
```
lovenunu: false ===

local/modules/Tinymce/Tinymce.php
```diff
@@ -51,11 +63,18 @@ public function postActivation(ConnectionInterface $con = null)
...
+                $fileSystem->mkdir($this->webMediaEnvPath . DS . 'upload');
+                $fileSystem->mkdir($this->webMediaEnvPath . DS . 'thumbs');
+            }
+        }else {
```
lovenunu: } else {

local/modules/Tinymce/Resources/js/tinymce/filemanager/config/config.php
```diff
@@ -69,47 +69,63 @@
 // Argh, url_site is not defined ?!
 if (empty($base_url)) {
     // A we did not used the router to access this dialog, we cannot use the URL class. Use the good old method.
-    $base_url = $request->getSchemeAndHttpHost() . preg_replace('!/tinymce/filemanager/dialog.php.*$!', '', $_SERVER['REQUEST_URI']);
+    $base_url = $request->getSchemeAndHttpHost().preg_replace('!/tinymce/filemanager/dialog.php.*$!', '',
+            $_SERVER['REQUEST_URI']);
```
lunika: why only this parameter on a new line ? Put all parameters on a new line or none.

local/modules/Tinymce/Tinymce.php
```diff
@@ -41,6 +43,16 @@ public function postActivation(ConnectionInterface $con = null)
     {
         $fileSystem = new Filesystem();
 
+        //Check for environment
+        if ($env = $this->getContainer()->getParameter('kernel.environment')) {
+            //Check for backward compatibility
+            if ($env !== "prod") {
```
lunika: this particular case doesn't cover default Thelia's behaviour. I can be in development mode and the folder ```media/dev``` will be created and never used. Moreover, the ```_dev``` part is never removed in the config file, so were files will be copied ?

lunika:
Once activated, if I change my environment value, it's not possible to upload a file anymore. I have this error message : 

```
There is an error! The upload folder there isn't. Check your config.php file. 
```

local/modules/Tinymce/Tinymce.php
Soldras: In the config.php we never add _dev to the environment so for exemple:

you have an environment TEST_THELIA, in dev mode you TEST_THELIA_dev but tinymce must not be dependent to dev / prod so it only use TEST_THELIA. So files will be copied in ```media/TEST_THELIA```

Soldras:
If you add an environment you have to activate the module for this environment ... It's the goal

lunika:
Well, in your particular case, not if you have a normal use of Thelia.

lunika:
Once a module activated, this module module is available for all the environments. In your case not because you have one database for each environment.

local/modules/Tinymce/Tinymce.php
lunika: ok I understand

lunika: but the dev folder is still created and never used :-)

bilhackmac:
It's useless to have multiples environnements with the same database :expressionless:

PR #1395, opened 2015-04-23
More explicit filenames for exports

air-dex:
Thelia's current default behavior for naming exports files (in the `Admin\ExportController`) consists in naming them "*data*", regardless of what the export consists in (customer, products, prices or other things depending on what should be exported). This PR consists in giving to the exports the possibility to have more explicit file names, depending on what they consist in.

Technically, the file name (without the extension) is now built in the `ExportHandler` instead of the formatter (where the "*data*" file name is a constant). It is now built in a method called `getFilename();` which returns the file name without its extension. The base `Thelia/ImportExport/Export/ExportHandler` class for ExportHandlers have a default `getFilename();` method which consists in naming the file "*data*", just like before in order to avoid breaking Thelia.

### DIFF
--- a/core/lib/Thelia/ImportExport/Export/ExportHandler.php
+++ b/core/lib/Thelia/ImportExport/Export/ExportHandler.php
@@ -29,6 +29,9 @@ use Thelia\ImportExport\AbstractHandler;
  */
 abstract class ExportHandler extends AbstractHandler
 {
+    /** @var string Default filename. */
+    const DEFAULT_FILENAME = "data";
+
     protected $locale;
 
     /** @var  array */
@@ -195,6 +198,16 @@ abstract class ExportHandler extends AbstractHandler
     public function isImageExport()
     {
         return $this->isImageExport;
+    }
+
+    /**
+     * Get filename
+     *
+     * @return string
+     */
+    public function getFilename()
+    {
+        return static::DEFAULT_FILENAME;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/lib/Thelia/ImportExport/Export/ExportHandler.php
```diff
@@ -198,6 +201,16 @@ public function isImageExport()
     }
 
     /**
+     * Get filename
+     *
+     * @return array
+     */
+    public function getFilename()
+    {
+        return $this::DEFAULT_FILENAME;
```
lovenunu: I'd rather use the ```static``` keyword instead of ```$this```. It's more explicit that we are accessing a constant dynamically

core/lib/Thelia/ImportExport/Export/ExportHandler.php
```diff
@@ -198,6 +201,16 @@ public function isImageExport()
     }
 
     /**
+     * Get filename
+     *
+     * @return array
```
lovenunu: Well ... I don't think this method returns an array

core/lib/Thelia/ImportExport/Export/ExportHandler.php
air-dex: Fixed: https://github.com/air-dex/thelia/commit/0fe6cbdce17538350d85eda0120add4da67b83a0

core/lib/Thelia/ImportExport/Export/ExportHandler.php
air-dex: Fixed: https://github.com/air-dex/thelia/commit/0fe6cbdce17538350d85eda0120add4da67b83a0

lovenunu:
Could you squash your commits please ?

lunika:
Thanks

PR #1396, opened 2015-04-23
Improved data encoding in exports

air-dex:
This PR deals with encoding data in exports.

In the `Admin/ExportController`, data to export are built and set in a `ImportExportEvent` for encoding purposes. Then the `TheliaEvents::EXPORT_BEFORE_ENCODE` event is dispatched. In this event, data to export might be altered. After that, Thelia should logically use data altered by the event instead of original data, but it uses original data. This PR consists in bringing back logic in Thelia, making it using data altered by the `TheliaEvents::EXPORT_BEFORE_ENCODE` event instead of original data.

### DIFF
--- a/core/lib/Thelia/Controller/Admin/ExportController.php
+++ b/core/lib/Thelia/Controller/Admin/ExportController.php
@@ -208,6 +208,7 @@ class ExportController extends BaseAdminController
                 ->setLang($lang)
             ;
 
+            $event->setData($data);
             $this->dispatch(TheliaEvents::EXPORT_BEFORE_ENCODE, $event);
 
             $formattedContent = $formatter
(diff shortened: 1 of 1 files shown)


lovenunu:
Well, as objects are all passed by reference in PHP, ```$data``` and ```Event::$data``` leads to the same memory address, so your PR doesn't change anything but adds two method calls.

PoC:
- Create the following listener on ```TheliaEvents::EXPORT_AFTER_ENCODE```:

```php
    public function foo(ImportExport $e)
    {
        $e->getData()->setData(["foo", "bar"]);
    }
```

- Then add a var_dump($data) after dispatching the event.
- Look at $data::$data value
- See that its value is ["foo", "bar"]
- Go make a coffee
- Ask your favorite search engine why
- Find an useful answer on stackoverflow.
- Drink your coffee
- Ask your favorite search engine why coffee helps you coding
- Find an answer on wikipedia
- Lock your session and go home

PR #1397, opened 2015-04-23
Column naming in CSV Exports more understandable for non-developer people

air-dex:
A Thelia current behavior in CSV exports consists in naming columns with aliases. These column names could not be suitable for non developer people who might be easier with explicit names, especially if the name is in a lang they understand. With this PR, it is now possible to give to the columns other names than aliases which might not be understood by some people.

Technically, a heading row with the column names is added to data to export through a `TheliaEvents::EXPORT_BEFORE_ENCODE` event. Then, datas are encoded and an unexpected heading row containing aliases is added to data to export. This unexpected heading row with aliases is then removed through the `TheliaEvents::EXPORT_AFTER_ENCODE` event. The methods used to do such a thing are located in `Thelia\Action\Export`, the Thelia's action dedicated to exports.

The header names could be translated too. It happens in the `Thelia/ImportExport/Export/ExportHandler` class. There is now a method called `getHeading();` which returns an array whose keys are aliases and whose values are i18n keys for names that should be displayed. There is also a `getTranslatedHeading();` method which returns an array with translated keys. As for the translation itself, the `Thelia/ImportExport/Export/ExportHandler` class now have a protected `Thelia\Core\Translation\Translator $translator` attribute and a `TRANSLATION_DOMAIN` constant containing the domain for translations. The `TRANSLATION_DOMAIN` constant can be overriden in extended classes and setting it to `null` disables translations. The `Thelia/ImportExport/Export/ExportHandler` class have a default behaviour which consists in using aliases. For this, it disables translations and uses headers whose names are aliases. This default behavior is also here in order to avoid breaking Thelia.

### DIFF
--- a/core/lib/Thelia/Action/Export.php
+++ b/core/lib/Thelia/Action/Export.php
@@ -75,15 +75,13 @@ class Export extends BaseAction implements EventSubscriberInterface
         $handler = $event->getHandler();
         $formatter = $event->getFormatter();
 
-        if ($handler instanceof ExportHandler &&
-            $formatter instanceof CSVFormatter
-        ) {
+        if ($formatter instanceof CSVFormatter) {
             // Get existing data
             $formatterData = $event->getData();
             $data = $formatterData->getData();
 
             // Get heading labels
-            $heading = $handler->getHeading();
+            $heading = $handler->getTranslatedHeading();
 
             // Complete heading row (with all keys)
             // - Use label if possible
--- a/core/lib/Thelia/Action/Export.php
+++ b/core/lib/Thelia/Action/Export.php
@@ -15,8 +15,11 @@ namespace Thelia\Action;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Thelia\Core\Event\Cache\CacheEvent;
+use Thelia\Core\Event\ImportExport;
 use Thelia\Core\Event\TheliaEvents;
 use Thelia\Core\Event\UpdatePositionEvent;
+use Thelia\Core\FileFormat\Formatting\Formatter\CSVFormatter;
+use Thelia\ImportExport\Export\ExportHandler;
 use Thelia\Model\ExportCategoryQuery;
 use Thelia\Model\ExportQuery;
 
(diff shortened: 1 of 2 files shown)


lovenunu:
Well ... If your code is CSV-centric, it should be in the csv formatter class, don't you think?

core/lib/Thelia/Action/Export.php
```diff
@@ -58,6 +61,77 @@ protected function cacheClear(EventDispatcherInterface $dispatcher)
...
+            $orderCols = array_keys($formatterData->getRow(0));
+            $orderCols = array_combine($orderCols, $orderCols);
+            $headingRow = array_merge($orderCols, $heading);
+
+            // Add heading row to the top of data
+            array_unshift($data, $headingRow);
+
+            // Update event's data
+            $formatterData->setData($data);
+            $event->setData($formatterData);
```
lovenunu: This line is useless. Object reference ... that kind of stuff

core/lib/Thelia/Action/Export.php
```diff
@@ -58,6 +61,77 @@ protected function cacheClear(EventDispatcherInterface $dispatcher)
...
+            $data = $formatterData->getData();
+
+            // Get heading labels
+            $heading = $handler->getTranslatedHeading();
+
+            // Complete heading row (with all keys)
+            // - Use label if possible
+            // - Use alias else
+            $orderCols = array_keys($formatterData->getRow(0));
+            $orderCols = array_combine($orderCols, $orderCols);
```
lovenunu: ```FormatterData::getRow``` may return false if the array is empty.
I'm not sure that ```array_combine``` really likes booleans :)

lovenunu:
Well ... In fact, we don't really need that, but in fact, it can be a module.
Moreover, you can do the same thing in the csv formatter without loosing the order.

If your code works, please do a module.
What do you think @bibich @roadster31 @lunika  ?

core/lib/Thelia/Action/Export.php
air-dex: > This line

You mean lines 97 and 98?

lovenunu: Nope, only 98.
On 97, $data is an array, so you only get a copy that you have to inject after modifications.

PR #1404, opened 2015-04-26
Revert patch and add 5.4 compatibility layer

lovenunu:
This PR removes @roadster31 patch #1399 and adds a compatibility layer that defines the array_column function for php 5.4.
This allows php >= 5.5 users to use the native implementation of this functions, while letting 5.4 users use this function.


lunika:
the ```core/composer.lock``` file is not needed any more. Remove it please.

lovenunu:
Done

PR #1422, opened 2015-05-05
New Smarty function : tcache

bibich:
New Simple Smarty function `tcache` to cache some parts of a template.

A key is generated with parameters passed to the function and allows to identify the uniqueness of the content. 
So you should keep attention to the parameters that are passed to the function.

# Parameters :

- **key** : (mandatory) a unique key
- **ttl** : (mandatory) a time to live in seconds
- **lang** : specific cache by lang, possible values : *on*, *off* (default: *on*)
- **currency** : specific cache by currency, possible values : *on*, *off* (default: *on*)
- **customer** : specific cache by customer, possible values : *on*, *off* (default: **off**)
- **admin** : specific cache by administrator, possible values : *on*, *off* (default: *off*)
- **role** : specific cache by role (none, customer, admin), possible values : *on*, *off* (default: *off*)
- **...** : you could (and should) add as many arguments as you need to generate a unique key and by the way prevents collisions

# Examples :

This example will cache the generate code of a navigation menu. It will generate different caches depending of the
current category, the lang and the currency. The cache will be used for 60 seconds, before being deleted :

```smarty
{tcache key="navigation" ttl="60" category={category attr="id"}}
<div class="navigation">
{loop name="category-navigation" type="category" }
....
<!-- other expensive loop -->
....
{/loop}
</div>
{/tcache}
```

Other examples : 

```smarty
{* cache for 10 minutes (600s) by lang, currency and customer *}
{tcache key="example" ttl="600" customer="on"}
...
{/tcache}

{* cache for 10 seconds by lang, currency and role (customer, admin or none) *}
{tcache key="example" ttl="600" role="on"}
...
{/tcache}

{* cache for 1 minute and without dependencies with lang and currency *}
{tcache key="example" ttl="60" lang="off" currency="off" }
...
{/tcache}
```


gillesbourgeat:
Hello,

Why not integrate the ability to easily change the cache driver since Thelia config ?
Example : Redis, Memcached

I will rather see a service using all drivers doctrine cache. Exemple 'thelia_cache'
The service retrieves the connection information and the pilot from the configurations Thelia
The smarty plugin uses the service 'thelia_cache'
This gives the possibility to the other developers to use the service 'thelia_cache'

Cheers,

lunika:
I agree with @gillesbourgeat it will be great to be able to use an other cache service.

And why you don't juste create a module ? It's just a simple smarty plugin. With a module, everybody can use it and don't have to upgrade to version 2.2 of Thelia.

gillesbourgeat:
it's possible to create a module that offers the service for thelia 2.0 and 2.1, and integrate the service in the core of version 2.2.
The name and functioning of the service should just stay the same.
Have the service in core allows optimizes thelia, example : cache config variables, sitemap, graphics in backoffice .....

bibich:
Yes I'm agree with you. I had started to implement what you suggest in an other PR, but it's pretty old.
So forget this PR, I will update the other one and make a module for older versions of Thelia.

PR #1424, opened 2015-05-07
Refactor & optimization of default template

mespeche:
1. Add Grunt & Bower configuration (src and dist assets)
	 - Src assets, contain Less files and non minified files.
	 - Dist assets, contain all minified files.
 2. Update Bootstrap & FontAwesome versions
 3. Separation of Thelia structure & Thelia theme
 4. Optimization of Less rules & selectors (put bootstrap classes directly into html instead of using theme in less files)
 5. Use only Bootstrap class type for Thelia components (invented buttons types no longer exist, like .btn-change-address...)
 6. Use cdn when exists and import files from assets if cdn doesn't respond
 7. Update Hooks html classes.

Css before :
- Size without Thelia rules : 131ko
- Size of Thelia rules : 129ko
- Full size : 260ko
- Number of selectors : 4973
- Number of rules : 2663

Css after :
- Size without Thelia : 132ko (update of Bootstrap & FontAwesome)
- Size of Thelia rules : 49ko
- Full size : 181ko
- Number of selectors : 3959
- Number of rules : 2390

Js before : 21ko
Js after : 8ko


lunika:
Can you create a new sale, activate it with some products to see the render ?

lunika:
I can't choose my payment solution

lunika:
In the footer, the subscribe button is cropped

lunika:
in the footer, links are no more inline

mespeche:
I fixed all the bugs you saw ;-)

Lucanis:
navbar-brand blocks are hidden in sm size

lunika:
I merge this PR, after that everybody can continue to work on this template with new PR.

Thanks

PR #1441, opened 2015-05-22
if rewrite url with /  this link not work

zzuutt:
if rewrite url with /  this link not work

### DIFF
--- a/templates/frontOffice/default/assets/src/js/thelia.js
+++ b/templates/frontOffice/default/assets/src/js/thelia.js
@@ -474,7 +474,7 @@ var pseManager = (function($){
                 $.ajax({type: "POST", data: $(this).serialize(), url: url_action,
                     success: function(data){
                         $(".cart-container").html($(data).html());
-                        $.ajax({url:"ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
+                        $.ajax({url:"/ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
                             success: function (data) {
                                 // Hide all currently active bootbox dialogs
                                 bootbox.hideAll();
--- a/templates/frontOffice/default/assets/src/js/thelia.js
+++ b/templates/frontOffice/default/assets/src/js/thelia.js
@@ -474,7 +474,7 @@ var pseManager = (function($){
                 $.ajax({type: "POST", data: $(this).serialize(), url: url_action,
                     success: function(data){
                         $(".cart-container").html($(data).html());
-                        $.ajax({url:"/ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
+                        $.ajax({url:url_site + "ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
                             success: function (data) {
                                 // Hide all currently active bootbox dialogs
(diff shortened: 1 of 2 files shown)


templates/frontOffice/default/assets/src/js/thelia.js
```diff
@@ -474,7 +474,7 @@ var pseManager = (function($){
                 $.ajax({type: "POST", data: $(this).serialize(), url: url_action,
                     success: function(data){
                         $(".cart-container").html($(data).html());
-                        $.ajax({url:"ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
+                        $.ajax({url:"/ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
                             success: function (data) {
```
roadster31: This is not working is Thelia is installed in a subdirectory (e.g. http://www.yourhost.com/shop).

You have to use `{url path="/ajax/addCartMessage"}` to get the right URL. Às thelia.js is not parsed by Smarty, you have to find a way to pass this URL, either as a form parameter, or form data attribute, or something.

templates/frontOffice/default/assets/src/js/thelia.js
```diff
@@ -474,7 +474,7 @@ var pseManager = (function($){
                 $.ajax({type: "POST", data: $(this).serialize(), url: url_action,
                     success: function(data){
                         $(".cart-container").html($(data).html());
-                        $.ajax({url:"ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
+                        $.ajax({url:url_site + "ajax/addCartMessage", data:{ product_id: product_id, pse_id: pse_id },
                             success: function (data) {
```
roadster31: This is not working, we cannot rely on url_site, which may be empty or wrong. Use a addCartMessage variable, which contains the URL (see below).

templates/frontOffice/default/layout.tpl
```diff
@@ -268,7 +268,13 @@ GNU General Public License : http://www.gnu.org/licenses/
     {block name="after-javascript-include"}{/block}
 
     {hook name="main.javascript-initialization"}
-
+    <script>
+        var url_site = '{config key="url_site"}';
+        if(url_site.charAt(url_site.length-1) != '/') {
+            url_site = url_site + '/';
+        }
+
```
roadster31: url_site is not reliable. Use a variable to hold the complete URL.

`var addCartMessage = "{url path='ajax/addCartMessage'}"`

zzuutt: ok I change the script accordingly

templates/frontOffice/default/layout.tpl
```diff
@@ -268,7 +268,11 @@ GNU General Public License : http://www.gnu.org/licenses/
     {block name="after-javascript-include"}{/block}
 
     {hook name="main.javascript-initialization"}
-
+    <script>
+       // fix path for addCartMessage
+       // if you use '/' in your URL rewriting, the cart message is not displayed
```
roadster31: A useful comment would be : "Define the addCartMessage url, used in thelia.js to update th mini-cart content.", or something meaningful.

roadster31:
Thanks :smiley:

Issue #1447, opened 2015-05-27
Coupon (1 product purchased / 1 offered)


zzuutt:
VAT the product offered is added to the total Price Taxed

![26-05-2015 15-51-04](https://cloud.githubusercontent.com/assets/7439945/7830437/9426f26a-044d-11e5-9420-1427f36e21bb.jpg)

PR #1448, opened 2015-05-27
apply discount on real taxed price for free product coupon

lunika:
for #1447


roadster31:
b8ffd2a5257278b214230b3f97999e1a

PR #1449, opened 2015-05-27
Add newsletter to customer loop and fix profile update form

lovenunu:
I wanted to know if my customer was registred to the newsletter, but ...
thuglife

Let's make it easier for everyone :)

### DIFF
--- a/core/lib/Thelia/Core/Template/Loop/Customer.php
+++ b/core/lib/Thelia/Core/Template/Loop/Customer.php
@@ -13,6 +13,7 @@
 namespace Thelia\Core\Template\Loop;
 
 use Propel\Runtime\ActiveQuery\Criteria;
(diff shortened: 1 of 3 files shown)


core/lib/Thelia/Core/Template/Loop/Customer.php
```diff
@@ -122,6 +126,24 @@ public function buildModelCriteria()
...
+
+        $search
+            ->addJoinObject($join)
+            ->withColumn("IF(ISNULL(".NewsletterTableMap::EMAIL."), 0, 1)", "is_registered_to_newsletter")
+        ;
+
+        if (false === $newsletter) {
+            $search->having("is_registered_to_newsletter = 0");
+        }
+
```
roadster31: What happened to true === $newsletter ?

Issue #1122, opened 2015-02-03
No way to display PSE image and documents in front-office


roadster31:
We have currently no way to display a Product Sale Element specific image and document in the front office.

A loop (or a parameter to image and document loop) is missing.

roadster31:
Solution: add product_sale_element parameter to image and document loop.

lovenunu:
Fixed with #1452

PR #1354, opened 2015-04-12
[WIP] PSE management revamping

roadster31:
The main goal of this PR is to offer a clean and responsive alternative to the bloated, table-based PSE edition system currently in use.

A visual feedback for attached images, documents and for the PSE's virtual document has been added.

The document and image loops have been improved (`pse` `parameter), to provide a way to access PSE attached image and documents, and fix #1122. 

The document loop also provide a way to get the virtual file attached to a PSE, througth the `virtual_product` argument.

The PR also fixes a problem which prevent Thelia core translations to be saved by the "Translation" feature.

Some screenshots of the new 'Price' tab :

img-2015-04-12 15 32 49
img-2015-04-12 15 32 36
img-2015-04-12 15 32 13

PR #1452, opened 2015-05-28
Add loops for pse image/document. Fix #1098 and #1122

### LINKED ISSUE #1098: Product combinations and associated images
In combination builder you can set a different image for every combination (for example a color), but when you visit the product page and choose for a specific combination, the product image does not change to the one you've setted in builder.
Neither you see the correct image in the cart preview or during the checkout process..
Is it supposed to work this way? any clue how to fix this bug?

lovenunu:


roadster31:
This PR is conflicting  #1354, which includes fixes for #1098 and #1122

gillesbourgeat:
Why not add a input parameter to the image or document loop ?

lovenunu:
Because #1354 already does it :)

gillesbourgeat:
@lovenunu :+1:

PR #1462, opened 2015-06-02
Change the default param of the method isPaid

Yochima:
By default the order paid status MUST BE OrderStatus::CODE_PAID only !


roadster31:
No.

By doing this, you're beaking backward compatibility.

PR #1463, opened 2015-06-02
Add argument type any list for loops

gillesbourgeat:
Hello,

Add argument type any list for loops

```php
Argument::createAnyListTypeArgument('my_param'),
```

Returns an array of strings

Cheers


lunika:
Thanks,

please create unit test for this new class.

gillesbourgeat:
it may be necessary to try prevent start or end spaces, or format the output values (trim())

PR #1474, opened 2015-06-05
Fix languages

enurit:


lunika:
you have conflicts with the following files : 

* core/lib/Thelia/Config/I18n/en_US.php
* core/lib/Thelia/Config/I18n/fr_FR.php
* templates/frontOffice/default/I18n/fr_FR.php

PR #1486, opened 2015-06-09
Fix `insert.sql` mistakes

gmichard:
- Fix incorrect mapping between `config` entries and their *i18n* entries
- Add `id` column insertion for `config` table to prevent 
  + about this mapping error possibility
  + **about an `AUTO_INCREMENT` value different of 1.** (without `config.id` specification : `config.id` and `config_i18n.id` couldn't match).
- Add missing default translation


lunika:
Thanks

Issue #1492, opened 2015-06-10
{format_date} format parameter inconsistency


roadster31:
When the `locale` parameter is not used, the accepted date format for the `format` parameter is PHP `date()` one.

When the (undocumented) `locale` parameter is used, the the accepted date format for the `format` parameter is PHP `strftime()` one.

PR #1494, opened 2015-06-11
Fix format_date function to use consistent format

bibich:
Fix for #1492 

Now the format is the same when attribute `locale` is used or not.
The format used is the same as `date` function : <http://php.net/manual/en/function.date.php>

ex.:

```smarty
{format_date timestamp={$smarty.now} format="D, d M Y H:i:s" locale="fr_FR"}
```


roadster31:
This PR causes a BC break, as templates which uses strftime format are broken.

Maybe you can check the format provided, and use the according date formating function ? if a % is present in the format, use strftime, otherwise use date()

bibich:
Yes, that's what I've done.

roadster31:
Sorry, I'd better read the code before complaining.

PR #1506, opened 2015-06-16
Fix index form error information session cleaning

bilhackmac:
Add `:<formType>` to index to remove from session error form


lunika:
Thanks

PR #1513, opened 2015-06-18
add 2 hooks to personalize order-products list in order-edit template

Yochima:
Same logic than the order list in backOffice


lunika:
As explained int the [documentation](http://doc.thelia.net/en/documentation/contribute.html#sql-scripts-modification) you can't modified the sql file directly.

PR #1528, opened 2015-06-25
Add date range for order export

Soldras:
* Add support for date and datetime type form field with option widget equal to choice in TheliaSmarty
* Add range date to export orders


core/lib/Thelia/Controller/Admin/ExportController.php
```diff
@@ -105,6 +105,14 @@ public function export($id)
                 );
             }
 
+            $periode = null;
```
lunika: periode is a french word

lunika:
You break the unit tests suite.

lunika:
:+1:

PR #1545, opened 2015-07-03
Define product clone route own id

bilhackmac:
Prevent clone route override create route


roadster31:
Grilled :  #1544 :stuck_out_tongue:

PR #1552, opened 2015-07-05
To solve/fix the problem &amp; in URL to only & when you click on pag…

Boyquotes:
…ination or on option select for choice the number of display  product on a page

See pagination or select here redirect to home page...
http://demo.thelia.net/?view=category&locale=en_US&category_id=1

### DIFF
--- a/templates/frontOffice/default/category.html
+++ b/templates/frontOffice/default/category.html
@@ -157,13 +157,13 @@
                                 </li>
                             {else}
                                 <li>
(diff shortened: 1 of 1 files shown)


lunika:
This seems needed only for version 2.1

lunika:
I just understand with it works on master branch.

In fact it is a bug that fixed this behaviour... I think we have to refactor the way we managed the "noamp" parameter.

lunika:
I fixed this problem in PR #1555

PR #1559, opened 2015-07-08
Added code, locale, and exclude default parameters

roadster31:
This PR add some useful parameters to the Lang loop.


lunika:
Thanks

PR #1561, opened 2015-07-08
Update templates to easierly handle order status created by modules

Mertiozys:
Template files (front and back) showing order status now use the status code for status created by modules, to create the CSS class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Add module image edition in backoffice
 - The language change links should now use the locale instead of the language code, e.g. http://www.yourshop/some-page?lang=fr_FR instead if http://www.yourshop/some-page?lang=fr. Backward compatibility is provided.
+- Order status added by modules have their CSS label color handled or have a default color
 
 # 2.2.0-beta1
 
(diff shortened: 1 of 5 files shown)


core/lib/Thelia/Config/I18n/en_US.php
```diff
@@ -55,7 +55,7 @@
     'Associate documents' => 'Associate documents',
     'Associate images' => 'Associate images',
     'At least one of cart products categories is %op% <strong>%categories_list%</strong>' => 'At least one of cart products categories is %op% <strong>%categories_list%</strong>',
-    'Attribut' => 'Attribut',
+    'Attribut' => 'Attribute',
```
bibich: It would be great to change the key too, in this file and where this key is used in translate functions.

core/lib/Thelia/Config/I18n/en_US.php
```diff
@@ -47,7 +47,7 @@
     'All Products' => 'All Products',
     'All countries' => 'All countries',
     'All shipping methods' => 'All shipping methods',
-    'An error happend while copying %prev to %dest' => 'An error happend while copying %prev to %dest',
+    'An error happend while copying %prev to %dest' => 'An error happened while copying %prev to %dest',
```
bibich: It would be great to change the key too, in this file and where this key is used in translate functions.

lunika: I don't agree, the key can't change. It doesn't matter if there is a typo, you can fixed it in the translation.

core/lib/Thelia/Config/I18n/en_US.php
lunika: I don't agree, the key can't change. It doesn't matter if there is a typo, you can fixed it in the translation.

bibich: Why ?? 
The replacement for other languages is not so difficult. And this key could be used as a fallback if a translation is not found.

bibich:
Thanks

PR #1563, opened 2015-07-10
fix ui in variables.html

gillesbourgeat:
before : 
variables thelia panel d addministration de thelia

after : 
variables thelia panel d administration de thelia


lovenunu:
Huuuu, you're merging your own PRs, what a disaster :scream:

PR #1564, opened 2015-07-11
fix i18n

gillesbourgeat:
fix i18n


templates/backOffice/default/product-edit.html
```diff
@@ -212,7 +212,7 @@
 
                         <div class="form-group">
                             {form_field form=$form field="newRef"}
-                            <label for="{$label_attr.for}">{intl l="{$label}"} </label>
+                            <label for="{$label_attr.for}">{intl l=$label} </label>
```
bibich: Should be just `{$label} ? It's already translated in the `ProductCloneForm`

PR #1567, opened 2015-07-12
Fix #1565 (Clone product, uncaught exception if reference already exists)

### LINKED ISSUE #1565: Clone product, uncaught exception if reference already exists
Clone product, uncaught exception if reference already exists

thelia loc index_dev php admin products clone

gillesbourgeat:
Fix #1565 (Clone product, uncaught exception if reference already exists)


bibich:
Thanks

PR #1569, opened 2015-07-13
fix #1568 hook position

### LINKED ISSUE #1568: Changing module order inside a hook is a pain
Changing module order inside a hook is a painfull process, as each time one's cliks on the up or down arrow, the page is reloaded and all the filter values (module name, hook name, etc.) are lost.

gillesbourgeat:
Fix hook position


bibich:
THanks

PR #1570, opened 2015-07-13
Language change fixes and improvements

roadster31:
This PR fixes a problem with langage change when the rewriting is enabled (lang is not changed in the 'lang' parameter appears in the URL).

The language change URL should now contain the locale (e.g. fr_FR) instead of the lang code (e.g. fr), to be sure to get the proper language. Until now, we cannot distinguish fa_FR from fr_CA.
A backward compatibility is provided.


bibich:
Thank you

PR #1574, opened 2015-07-16
Prevent short research and keep research in input

gillesbourgeat:
Prevent short research and keep research in input


bibich:
danke

PR #1580, opened 2015-07-20
WIP  Fix and close #1577

### LINKED ISSUE #1577: Log in Thelia
Since the version 2.1.4, we have a problem with all module logs.

To understand the problem, check the log of any Thelia module (with its own log file). They are generated in the wrong path and in the wrong file.

The problem come from the new function :
protected function findRelativePath($filePath, $mode)

This function is called twice by all the Thelia module using a specific log-file.

Thx

gillesbourgeat:
Fix and close #1577


gillesbourgeat:
This solution is not compatible with windows

PR #1581, opened 2015-07-20
Fix the prices precision

gillesbourgeat:
Fix the prices precision

Example of price ttc problematic before : 
- 59.90
- 308.00
- 20000.00


bibich:
Thanks

PR #1588, opened 2015-07-23
Adds the document tab in product page

gillesbourgeat:
Adds the document tab in product page

fr_fr fr_fr tu texag fr_fr trois heures thelia v2


bibich:
Thanks

PR #1592, opened 2015-07-24
fix InvalidArgumentException in Form.php line 586

zzuutt:
InvalidArgumentException in Form.php line 586: Missing 'form' parameter in form arguments

forgot form parameter


gillesbourgeat:
Thanks !

PR #1593, opened 2015-07-24
Remove any UploadedFile object from session form data

bilhackmac:
UploadedFile objects are not serializable and makes php-cgi process to freeze


gillesbourgeat:
Thanks !

PR #1598, opened 2015-07-27
Update faker to stable version 1.5

gillesbourgeat:
Update faker to stable version 1.5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - #1590 Fix Thelia request initialization
 - #1593 Fix form serialization in session that contain uploaded files
 - update symfony/validator version to 2.3.31
+- #1598 composer.json update dependency fzaninotto/faker to stable version 1.5
 
 # 2.2.0-beta2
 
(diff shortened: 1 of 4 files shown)


lovenunu:
You forgot to update ```setup/composer.json```, and I think you didn't update the ```composer.lock``` too :)

gillesbourgeat:
@lovenunu :+1:

lunika:
If you want to update a library, use composer require with the required version instead of updating manually the composer.json file

lunika:
I ignore how you update the faker but not only the faker was updated. Take a look at the diff on composer.lock file.

smarty's source has been changed from git to svn but the svn repo for this version does not exists anymore.

Revert all your modification and simply run : 

```
composer require fzaninotto/faker:1.5.*
```

or

```
composer require fzaninotto/faker:~1.5
```

gillesbourgeat:
Yes, I used this commands.
From a new local instance of the thelia repository, I did not have this problem.
Yet, my old local instance is up to date.
Thanks !!!

PR #1601, opened 2015-07-28
Add set error in TheliaFormValidator when form is not valid

Alban-io:
Currently [Thelia BaseForm has_error](https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Form/BaseForm.php#L80) must be set manually or with setErrorMessage().
I think this is TheliaFormValidator responsibility to set this value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 2.2.0-beta3
 
+- #1601 Add set error in TheliaFormValidator when form is not valid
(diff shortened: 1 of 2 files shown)


gillesbourgeat:
Can you update the changelog please, and squash your commits.

Alban-io:
done :smile:

gillesbourgeat:
Thanks !

PR #1611, opened 2015-08-04
Prevent obscur "[] this value cannot be null" messages.

roadster31:
The field name is always used instead an empty ('') field label.


gillesbourgeat:
:+1:

bibich:
Thanks

PR #1613, opened 2015-08-04
Fix fieldVars param in renderFormCollectionField smarty block

Alban-io:
Currently in the [Form::renderFormCollectionField()](https://github.com/thelia/thelia/blob/master/local/modules/TheliaSmarty/Template/Plugins/Form.php#L876) method the assignFieldValues $fieldVars param is the FormConfigBuilder options attribute.

I replace it with the current FormView field vars like in [Form::processFormField()](https://github.com/thelia/thelia/blob/master/local/modules/TheliaSmarty/Template/Plugins/Form.php#L321) because the FormConfigBuilder options doesn't contains the Form error.


lovenunu:
wow

Alban-io:
:confused: ?

bibich:
faleminderit

PR #1614, opened 2015-08-04
Product loop ref argument is now a list

roadster31:
This argument is already processed as a list in the loop code.


bibich:
Thanks

PR #1615, opened 2015-08-05
New TheliaEvents::CART_FINDITEM event to improve cart management flexibility

roadster31:
This PR introduce a little bit more flexibility in cart management.

Until now, the `Cart::findItem()` method was called to check if a cart item with the same PSE already exists in the current cart.

However, some modules may want to change the way the Cart action is searching for existing items, but they can't, as `Cart::findItem()` is a protected method.

This PR introduces the `TheliaEvents::CART_FINDITEM` event, which is now used in the Cart action instead of a direct call to `Cart::findItem()`, allowing modules to perform some specific actions.

The `Cart::findItem()` method still exists to ensure a backward compatibility, but is deprecated.

### DIFF
--- a/core/lib/Thelia/Action/Cart.php
+++ b/core/lib/Thelia/Action/Cart.php
@@ -278,13 +278,13 @@ class Cart extends BaseAction implements EventSubscriberInterface
      */
     public function findCartItem(CartEvent $event)
     {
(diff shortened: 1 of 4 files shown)


bibich:
Thank you

Issue #899, opened 2014-11-19
Category menu behavior is not correct


m190:
Category menu behavior is not correct for categories with more than 1 depth. 
If is chosen for example the third level, the menu appears collapsed. 
Also there is no way to choose second level if you have subcategory (third level), after clicking on it you just only rolls up. 
The view of collapse/roll up icon is also is not correct for deeper levels.

PR #1619, opened 2015-08-06
Fix issue #899 : category menu collapse behavior & update bootsrap

Mertiozys:
Category side menu now collapse well :
- subcategories are clickable
- collapse buttons display the good icon all the time


bibich:
Thanks

PR #1626, opened 2015-08-11
Fix form in session

### LINKED ISSUE #1597: Setting form data in session breaks coupon
Hi,

I'm working currently in Thelia beta 2.2.

When i render a coupon update template (/admin/coupon/update/{$id}) my php session is broken because :

In the [ParserContext::addForm() method](https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Core/Template/ParserContext.php#L124), when adding form data in session there is a problem because the Coupon condition classes all extends [ConditionAbstract](https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Condition/Implementation/ConditionAbstract.php#L42) which contains [BaseFacade](https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Coupon/BaseFacade.php#L40) which contains application container (this object is unserializable and can not be serialized) !

gillesbourgeat:
This pull request offers a solution to avoid passing all forms in the session
The forms is added to the session only if the error url is present

resolve #1597

### DIFF
--- a/core/lib/Thelia/Core/Template/ParserContext.php
+++ b/core/lib/Thelia/Core/Template/ParserContext.php
@@ -31,13 +31,19 @@ class ParserContext implements \IteratorAggregate
     // Lifetime, in seconds, of form error data
     const FORM_ERROR_LIFETIME_SECONDS = 60;
 
+    /** @var BaseForm[] */
     private $formStore = array();
+
+    /** @var array */
     private $store = array();
 
-    private $formFactory ;
+    /** @var TheliaFormFactoryInterface */
+    private $formFactory;
+
+    /** @var TheliaFormValidatorInterface */
     private $formValidator;
 
-    /** @var  Request */
+    /** @var Request */
     private $request;
 
     public function __construct(
@@ -72,6 +78,7 @@ class ParserContext implements \IteratorAggregate
     /**
      * Set the current form.
      *
+     * @param null $default
      * @return BaseForm|null
      */
     public function popCurrentForm($default = null)
@@ -85,6 +92,10 @@ class ParserContext implements \IteratorAggregate
         return $form;
     }
 
+    /**
+     * @return BaseForm
+     * @throws SmartyPluginException
+     */
     public function getCurrentForm()
     {
         $form = end($this->formStore);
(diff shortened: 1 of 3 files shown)


Alban-io:
:+1:

core/lib/Thelia/Core/Template/ParserContext.php
```diff
@@ -108,29 +119,31 @@ public function getCurrentForm()
...
-        $formErrorInformation = $this->request->getSession()->getFormErrorInformation();
-
         $this->set(get_class($form) . ":" . $form->getType(), $form);
 
-        /** Remove unserializable objects \Symfony\Component\HttpFoundation\File\UploadedFile */
-        $formData = $form->getForm()->getData();
-        foreach ($formData as $idx => $value) {
-            if ($value instanceof UploadedFile) {
-                unset($formData[$idx]);
+        if ($form->getForm()->get("error_url")->getData() !== null) {
```
roadster31: Why don't you use BaseForm::getErrorUrl() ?

roadster31:
This PR is just a workaround, and is not addressing our main concern about forms : defining a standard way to process form errors (specially in front office), and designing a clean and robust form lifecycle, which includes form errors feedback to the user.

lunika:
:+1: with roadster

roadster31:
Additionnaly, the PR is breaking compatibility with situations where error_url is not defined, while the form stored in session is nevertheless used by the   displayed template

gillesbourgeat:
I know that this pr is just a workaround. But it is not possible to pass automatically all forms in the session, for use of this behavior very seldom.
I'm sorry, but I not see a use case for this behavior.
However, it's possible to add a parameter "pushInSeesion" on the method ```addForm```
```php
// if $pushInSeesion is true or error_url is not empty, then add the form in session.
public function addForm(BaseForm $form, $pushInSeesion = false)
```

core/lib/Thelia/Core/Template/ParserContext.php
gillesbourgeat: Indeed !

roadster31:
In the back-office, most error handling is performed by redisplaying the current template, without redirecting, thus the form remains in the current context, and a session storage is not required.

However, many modules do not use this system, and if you give a look, you'll see that most of them, simply don't process form errors, because there's no way to do it !

The situation is the same in the front office.

Thus, IMO, this PR is not an "enhancement", but just a "temporary workaround". Additionnaly, we *should not* introduce parameters such as $pushInSeesion, which are only a way to partially compensate a major form lifecycle design flaw.

We should invest some time to work on form error management.

PR #1627, opened 2015-08-11
WIP : Auto inject form in hook

gillesbourgeat:
This pull request adds automatically the parameter "form" if the hook is in a form
close #1325


gillesbourgeat:
close #1325

PR #1630, opened 2015-08-12
[setup] add parent folder

andrey1s:


bibich:
`0` is already the default value for parent in `folder` so I don't really understand your PR ?
So I will close this PR.

PR #1632, opened 2015-08-13
Resolve #1623 : pagination & limit issues

### LINKED ISSUE #1623: Pagination Next Prev Limit issue
Hi,

I noticed some issues in pagination. The first one is the Next - Prev issue: if you try to go to the second page in category, and then back to first page, and repeat this some more time and you will get an URL like this: http://localhost/thelia-trunk/web/index_dev.php/?category_id=1&limit=12&locale=en_US&page=1&page=2&page=2&view=category&page=1

The second issue is when you are on send page, but you choose such amount of items per page, which will could be shown only on one page. Then you will get a page with warning no items. The correct behavior should be redirect on the prev page, or not redirect, but correct URL generation from the beginning.

Thanks,

Alexandr

Mertiozys:
Fix stacking parameters in URL when using pagination.
Redirect on page=1 when changing the limit of product per page to avoid being on a page with no products.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - #1634 Remove leftover uncallable routes (admin)
 - #1631 Remove duplicate route (admin)
 - #1629 Fix errors reporting of admin hooks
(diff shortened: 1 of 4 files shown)


bibich:
Thanks

PR #1634, opened 2015-08-13
Remove leftover uncallable routes

bcabarrou:
The controller methods for these routes were removed. As far as I can tell, they are not used.


gillesbourgeat:
Thanks !

PR #1635, opened 2015-08-13
Add routing files checks to unit tests

bcabarrou:
Check the core and front routing files for:
- Duplicate route ids
- Duplicate route paths (not an exact test)
- Uncallable controller methods
- Non-existent templates (for front routes)

Right now this PR will break the tests since there were in fact some problems :) They are fixed in PRs  #1631 and #1634.

### DIFF
--- a/tests/phpunit/Thelia/Tests/Config/RoutesConfigTest.php
+++ b/tests/phpunit/Thelia/Tests/Config/RoutesConfigTest.php
@@ -1,0 +1,282 @@
+<?php
+/*************************************************************************************/
+/*      This file is part of the Thelia package.                                     */
+/*                                                                                   */
+/*      Copyright (c) OpenStudio                                                     */
+/*      email : dev@thelia.net                                                       */
+/*      web : http://www.thelia.net                                                  */
+/*                                                                                   */
+/*      For the full copyright and license information, please view the LICENSE.txt  */
+/*      file that was distributed with this source code.                             */
+/*************************************************************************************/
+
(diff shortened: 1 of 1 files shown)


lunika:
Upper case variables are reserved for constant. Please just use camel case like anyone variable.

bcabarrou:
Fixed.

bcabarrou:
Added license header.

gillesbourgeat:
Thanks !

lunika:
Why did you use ```static``` instead of ```self``` in this case?

bcabarrou:
I used `self` for writing properties, `static` to read them. That way if the class is extended, they can be redefined and affect the base methods (in that case we probably want that).

Is there a possible issue with that ?

lunika:
Nop I just wanted to know and your explanation is clear. 

Thank you

lunika:
But I'm not sure it will works as you expect. They will have to override the part where you write in the static variable because with the self instruction, it takes the context where it was defined and not where it was called.

bcabarrou:
You are right, just tested it: https://gist.github.com/bcabarrou/61cb9da59ee5dc119127

Reading with `static` will only read a child class static attribute if the child class explicitly define that attribute. I though that static attributes were inherited as 'separate' to the parent class, but that is not the case (which I guess makes sense when you think about it: static attributes are a property of the *class* itself).

So yeah in that case it still works, but you have to explicitly redefine the static attributes you want to override (even if you initialize them elsewhere - just re-attributing them a value will not work).

PR #1637, opened 2015-08-13
Resolve #1636 : impossibility to edit an API. Fix some CSS stuff.

### LINKED ISSUE #1636: In the backoffice it is not possible to edit api
In the backoffice it is not possible to edit api

path : /admin/configuration/api

Mertiozys:
API edit button now opens the edition modal when clicked.
"Add" button for attribute and feature values doesn't overflow on the blue info panel.


gillesbourgeat:
Can you squash your commits.

PR #1641, opened 2015-08-18
[setup] fix faker real text

andrey1s:


gillesbourgeat:
This pull request cannot be merged, that changes nothing.

roadster31:
Using mb_strtolower() is a good idea.

PR #1643, opened 2015-08-18
Resolve #1280 : stock problem when updating order status

### LINKED ISSUE #1280: Payment module + Stock
if we add the function
```    public function manageStockOnCreation()
    {
        return false;
    }```
in the module payment.and the order is not paid
in back side, when cancels the order the products are added to the stock.

Normally the stock should be added only if the order is paid and canceled

Mertiozys:
Stocks are now well managed when changing an order status.
Error message if no stock is now translated.
This solves the problem but is not very evolutive, if some status are added they won't be taken in charge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 2.2.0-beta3
 
+- #1643 Fix stock errors when changing order status
 - #1637 Fix admin API edit button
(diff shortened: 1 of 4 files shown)


bibich:
What about the status of this PR ? is it still valid ?

PR #1646, opened 2015-08-20
don't store form object in session

lunika:
As explained in #1597 storing object in session is not a good practice, we have issues with the serialization of some objects.

I suggest to never store object in session but keep all existing data who are not objects.

What do you think ?


gillesbourgeat:
This solves the problem for version 2.2 and gives the opportunity to find a better solution later.

I suggest to add a recursive behavior for a collection data

```php
    protected function cleanFormData($data)
    {
        foreach ($data as $key => $value) {
            if (is_array($value)) {
                $data[$key] = $this->cleanFormData($value);
            } elseif (is_object($value)) {
                unset($data[$key]);
            }
        }

        return $data;
    }

   public function addForm(BaseForm $form)
    {
        $formErrorInformation = $this->request->getSession()->getFormErrorInformation();

        $this->set(get_class($form) . ":" . $form->getType(), $form);

        $formData = $this->cleanFormData($form->getForm()->getData());

       ......
```

ping @roadster31

PR #1652, opened 2015-08-24
Don't store form object in session

### LINKED ISSUE #1597: Setting form data in session breaks coupon

gillesbourgeat:
This pull request follows #1646

Close #1646
Resolve #1597

### DIFF
--- a/core/lib/Thelia/Controller/BaseController.php
+++ b/core/lib/Thelia/Controller/BaseController.php
@@ -323,7 +323,7 @@ abstract class BaseController extends ContainerAware
         $url = null;
 
         if ($form != null) {
-            $url = $form->getFormDefinedUrl($parameterName, $form);
+            $url = $form->getFormDefinedUrl($parameterName);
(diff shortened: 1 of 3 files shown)


bibich:
Thanks

PR #1653, opened 2015-08-24
Deprecated admin includes

gillesbourgeat:
Hello,
While it was pretty useful for some time, the AdminInclude functionnality is now deprecated.

The question is, we must remove the admin includes in version 2.3 or 2.4 ?

What do you think ?


roadster31:
Why deprecating it ?

gillesbourgeat:
- it was planned to deprecate the admin includes
- they duplicate the hook behavior
- they are few
- It's difficult to have a overload behavior

PR #1656, opened 2015-08-26
PSE ref don't change if we change product ref

InformatiqueProg:


core/lib/Thelia/Action/Product.php
```diff
@@ -110,7 +112,14 @@ public function update(ProductUpdateEvent $event)
                     ->save($con)
                 ;
 
-                // Update default category (ifd required)
+                // Update default PSE (if product has no attributes and the product's ref change)
+                if ($prev_ref != $product->getRef() && $product->countSaleElements() == 1) {
```
Mertiozys: This checks if there is 1 PSE, not if there are no attributes
What if there is 1 attribute combination ? :-)
You should add a condition to your test if $product->getDefaultSaleElements()->countAttributeCombinations() == 0

core/lib/Thelia/Action/Product.php
```diff
@@ -95,6 +95,8 @@ public function update(ProductUpdateEvent $event)
             $con->beginTransaction();
 
             try {
+                $prev_ref = $product->getRef();
```
Mertiozys: Please use camelCase instead of underscore_case

core/lib/Thelia/Action/Product.php
InformatiqueProg: Create a product with 1 attribute combination?  :'-(
After all why not

Mertiozys: Yes that's weird but this case has to be taken into account :-)

bibich:
Hi.
#1790 fixed this issue. 
So I close your PR.
Thanks.

Issue #1666, opened 2015-09-04
No error is displayed when trying to delete a customer which has orders


roadster31:
It is not allowed to delete a customer which has orders. However, the basket icon is displayed in the customer tool menu :

![img-2015-09-04 23 51 51](https://cloud.githubusercontent.com/assets/2197734/9695240/071fc72c-5360-11e5-8e3a-b354eabc5385.png)

If one's try to delete this customer, the customer is not deleted (this is OK), but no error message is displayed.

gillesbourgeat:
Hello and Thanks !

Do you have found a solution for the position of the currency symbol, as indicated on your image ?
The PR #1667 does not solve the problem.

roadster31:
No, I did not address the currency symbol issue in #1667. In fact, I did not realized the symbol was at the wrong place.

PR #1667, opened 2015-09-04
Fix for #1666

roadster31:
A question mark with a tooltip replace the trash icon for customers with orders.

img-2015-09-05 00 07 53


gillesbourgeat:
Thanks !!

Resolve #1666

PR #1669, opened 2015-09-08
Order loop improvements

roadster31:
This PR adds three new parameters to the order loop:
- exclude_status, to exclude some status IDs
- status_code, to filter using a status code (paid, not_paid, processing, etc.)
- exclude_status_code, to exclude some status codes.

A new output variable is added, to return the status code (paid, not_paid, etc.) : 
- $STATUS_CODE


gillesbourgeat:
Thanks !

PR #1676, opened 2015-09-11
Require required PHP extensions with composer [WIP]

bcbrr:
Since Thelia depends on a few PHP extensions, it might be a good idea to formalize it.

(see https://getcomposer.org/doc/02-libraries.md#platform-packages)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,12 @@
     },
     "require": {
         "php": ">=5.4",
+        "ext-pdo_mysql": "*",
+        "ext-mcrypt": "*",
+        "ext-intl": "*",
+        "ext-gd": "*",
+        "ext-curl": "*",
+        "ext-calendar": "*",
         "ircmaxell/password-compat": "1.0.*",
         "propel/propel": "dev-master#793aea6b762d9a72b083bfc49c74c574f67650c8",
         "psr/log": "1.0",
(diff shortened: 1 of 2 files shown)


gillesbourgeat:
:+1:

Alban-io:
:+1:

composer.json
```diff
@@ -15,6 +15,12 @@
     },
     "require": {
         "php": ">=5.4",
+        "ext-pdo_mysql": "*",
+        "ext-mcrypt": "*",
+        "ext-intl": "*",
+        "ext-gd": "*",
```
lunika: Thelia can be used without gd but with imagine. This verification is also wrong in the CheckPermission class

bibich: Yes, you can use gd / imagick or gmagick but in composer you can not have conditionnal requirements, so I think you have to remove this requirement.
But it would be great to add this check in the CheckPermission class.

bcabarrou: Ok, I will look into this.

lovenunu: You can add suggestions that will be displayed to the user after the install :)

lunika: or a script using the `pre-install-cmd` hook https://getcomposer.org/doc/articles/scripts.md

lopes-vincent:
Outdated PR

lunika:
> Outdated PR

Time is something relative.

Issue #1678, opened 2015-09-12
attribute_combination don't apply attribute 's template order


InformatiqueProg:
On product page the PSE title, on "ajax add cart", on cart

![capture](https://cloud.githubusercontent.com/assets/7933327/9833600/bfde5d22-599c-11e5-8825-f4d92d0f00ca.png)

PS: commencer par la tarte au citron, je veux bien, mais passer au cambert après je dis NON :worried:

InformatiqueProg:
Same problem with OrderProductAttributeCombination loop,
I can push the change in this PR or you need new one ?

PR #1679, opened 2015-09-12
Add manual order in attributeCombination loop

InformatiqueProg:
see issue #1678


bibich:
Thanks but this is not really a bug. So I will not merge your PR for version 2.1 
But you could clean it a little and open a PR on the master branch.

For the loop I think that you could use something like this instead : 

```php
...
                case "manuel":
                    $search
                        ->joinAttribute()
                        ->addAscendingOrderByColumn(AttributeTableMap::POSITION)
                    ;
                    break;
                case "manuel_reverse":
                    $search
                        ->joinAttribute()
                        ->addDescendingOrderByColumn(AttributeTableMap::POSITION)
                    ;
                    break;
...
```

Issue #1689, opened 2015-09-18
Exception when forwarding a request to a controller


bcabarrou:
Forwarding a request to another controller action (sometimes ?) causes this exception:

```
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: The scope "request" is not active. in (...)/vendor/symfony/dependency-injection/Symfony/Component/DependencyInjection/Container.php on line 459
```

PR #1690, opened 2015-09-18
Fix HTTP sub-requests causing two container request scope exit

bcabarrou:
The type of the request (master or sub) is not checked when entering and leaving the container request scope. Thus, sub-requests (such as the one created by `BaseAdminController::forward()`) causes multiple exits from the scope, raising an exception.

Any possible side-effects from this change ?

This fixes https://github.com/thelia/thelia/issues/1689.


bibich:
Thanks

PR #1692, opened 2015-09-21
Version 2.2.0

### LINKED ISSUE #1666: No error is displayed when trying to delete a customer which has orders

### LINKED ISSUE #1693: Order postage's tax don't respect logical tax rule but "order_product" do
In order_product, prices are save without tax but postage's price is with Tax...

And rounding problems are back !

bibich:
- Fix amounts displayed on the PDF invoice when a postage with tax is used (fixes #1693 and #1694)
- Fix translations for HookNavigation module
- Update hooktest-template and hooktest-module to prevent thelia-installer conflicts
- Update French, German, Italian translations
- Add Turkish translation
- #1688 Fix the permission messages in Thelia installer
- #1686 Use createForm method for front forms ```thelia.coupon.code, thelia.order.delivery, thelia.order.payment```
- #1667 Fix #1666 Display an error when trying to delete a customer which has orders
- #1665 Fix form field type date in Smarty plugin form, checks if the field type is a BirthdayType for assign a smarty variable [years, month, days]
- #1659 Fix Administrator edit action in the BackOffice, it was impossible to edit an administrator


roadster31:
:+1:

PR #1701, opened 2015-09-24
Order::createOrder() improvement to allow order duplication.

roadster31:
This PR improves the `Order::createOrder()` so that the method could be used to duplicate an order by re-using the delivery and invoice addresses defined in the original order.

The `$useOrderDefinedAddresses` parameter has been  added to select the required behaviour. If `true`, the order addresses are re-used. If `false`, the getChoosen*XXXX*Address() methods are used to get the customer addresses choice, and new `OrderAdress`es are created.

### DIFF
--- a/core/lib/Thelia/Action/Order.php
+++ b/core/lib/Thelia/Action/Order.php
@@ -382,6 +382,7 @@ class Order extends BaseAction implements EventSubscriberInterface
 
     /**
      * Create an order outside of the front-office context, e.g. manually from the back-office.
+     * @param OrderManualEvent $event
      */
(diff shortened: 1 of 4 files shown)


bibich:
Thanks. 
It would be great to update unit tests. 
In fact, there is not yet tests for `createManual` action, so could you implement some ?

roadster31:
I'd like to add test for createManual action, but have no time for this at the moment.

roadster31:
Is something missing o this PR (apart unit tests) ?

roadster31:
The PR is now complete.

bibich:
Thanks

PR #1702, opened 2015-09-25
WIP : Image, other sources

gillesbourgeat:
Hello,

Add the ability to have other sources for images.

Add table image
```xml
  <table name="image" namespace="Thelia\Model">
    <column autoIncrement="true" name="id" primaryKey="true" required="true" type="INTEGER" />
    <column name="source" required="true" size="50" type="VARCHAR" />
    <column name="source_id" required="true" type="INTEGER" />
    <column name="file" required="true" size="255" type="VARCHAR" />
    <column defaultValue="1" name="visible" required="true" type="TINYINT" />
    <column name="position" type="INTEGER" />
    <column name="title" size="255" type="VARCHAR" />
    <column name="description" type="CLOB" />
    <column name="chapo" type="LONGVARCHAR" />
    <column name="postscriptum" type="LONGVARCHAR" />
    <index name="idx_image_source_source_id">
      <index-column name="source" />
      <index-column name="source_id" />
    </index>
    <index name="idx_image_source_source_id_position">
      <index-column name="source" />
      <index-column name="source_id" />
      <index-column name="position" />
    </index>
    <behavior name="validate">
      <parameter name="sourceRule" value="{column: source, validator: Regex, options: {pattern: &quot;/[a-z0-9\-]{3,50}/&quot;}}" />
    </behavior>
    <behavior name="timestampable" />
    <behavior name="i18n">
      <parameter name="i18n_columns" value="title, description, chapo, postscriptum" />
    </behavior>
  </table>
```
The name of the source is limited to 50 characters, only numbers, letters and dashes.

Loop image example
```smarty
{loop type="image" name="image" source="my-source" source_id=44}
```
The old behaviors of the image loop is not changed.

Cheers,

### DIFF
--- a/core/lib/Thelia/Core/Template/Loop/Image.php
+++ b/core/lib/Thelia/Core/Template/Loop/Image.php
@@ -19,6 +19,7 @@ use Thelia\Core\Template\Loop\Argument\Argument;
 use Thelia\Core\Event\Image\ImageEvent;
 use Thelia\Core\Event\TheliaEvents;
 use Thelia\Core\Template\Loop\Argument\ArgumentCollection;
+use Thelia\Model\ImageQuery;
 use Thelia\Model\ProductDocumentQuery;
 use Thelia\Model\ProductImage;
 use Thelia\Type\BooleanOrBothType;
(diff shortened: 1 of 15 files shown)


roadster31:
There is a problem with this solution : deleting the "source" (product, content, folder, whatever) will not cascade, and we have to care about deleting the related images (and documents ?) entries in the application code.

roadster31:
> {pattern: &quot;/[a-z0-9\-]{3,50}/&quot;}}

why {3,50} ? what about underscores ? This rule is too restrictive. Checking for a valid identifier name would be better :

`identifier ::=  (letter|"_") (letter | digit | "_")*`

Thus something like `^[^\d\W]\w*$` would be fine;

PR #1714, opened 2015-10-05
Hook improvements

bibich:
This PR aims to simplify the use of hooks and fixes some issues.

Now, it's no longer required to implement your own class to handle hooks. If you only want to render a template or add a css or js file you can do it easily. Here is an example : 

```xml
<hooks>
    <hook id="hooknavigation.hook.test">
      <tag name="hook.event_listener" event="main.navbar-primary" templates="render:myTemplate.html" />
      <tag name="hook.event_listener" event="main.stylesheet" type="front" active="1" templates="css:myCss.css;render:theme.css.html" />
    </hook>
</hooks>
```

You can notice that the `class` argument has been omitted as well as the `scope` argument.

The tag syntax is similar and it operates the same way,  except for the new argument `templates`. This argument is used to define what contents are going to be *injected* in the hook. 

The syntax : 

```
templates="<action>:<file>[;<action>:<file>]*"
```

- action (what to do with the file) :
    - **render** : this is the default one (if action is omitted). It will render the file. all parameters affected to the hook will be available in the template.
    - **dump** : will just dump the content of the file.
    - **css** : will create a `link` html tag and add the CSS file.
    - **js** : will create a `script` html tag and add the JavaScript file.
- file : this is the full path of the file related to the root directory of your template directory. For example, if the hook is a front office hook and want to render the file `myTemplate.html`. This template should be located in your module inside the `templates\frontOffice\default\` directory. The overriding system is the same as used with classic hook.

I've also added a command to cleanup hooks of a module or for all modules. They will be recreated (with default configuration) when the cache of the application will be cleared :

```bash
# remove module hooks of the HookNavigation module (without confirmation)
php Thelia hook:clean HookNavigation --assume-yes
# remove hooks of all modules (with confirmation)
php Thelia hook:clean
```

Feel free to comment this PR. It's not finished and all your remarks are welcome.

### DIFF
--- a/core/lib/Thelia/Action/ModuleHook.php
+++ b/core/lib/Thelia/Action/ModuleHook.php
@@ -117,6 +117,7 @@ class ModuleHook extends BaseAction implements EventSubscriberInterface
             ->setModuleActive($this->isModuleActive($event->getModuleId()))
             ->setHookActive($this->isHookActive($event->getHookId()))
             ->setPosition($this->getLastPositionInHook($event->getHookId()))
+            ->setTemplates($event->getTemplates())
             ->save();
 
         // Be sure to delete this module hook from the ignored module hook table
@@ -139,6 +140,7 @@ class ModuleHook extends BaseAction implements EventSubscriberInterface
                 ->setMethod($event->getMethod())
                 ->setActive($event->getActive())
                 ->setHookActive($this->isHookActive($event->getHookId()))
+                ->setTemplates($event->getTemplates())
                 ->save();
 
             $event->setModuleHook($moduleHook);
--- a/core/lib/Thelia/Command/HookCleanCommand.php
+++ b/core/lib/Thelia/Command/HookCleanCommand.php
@@ -1,0 +1,153 @@
+<?php
+/*************************************************************************************/
+/*      This file is part of the Thelia package.                                     */
+/*                                                                                   */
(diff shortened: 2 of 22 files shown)


lunika:
I don't like this :-1:

roadster31:
What is wrong for you @lunika ?

Providing shortcuts to simplify hooks management and reduce the amount of code is a nice idea.

lunika:
it obfuscates how `hook` work, why it is interesting to declare all this amount of class and php. One of the good thing with hooks is that you can remove the use of `loop` and have a more traditional `MVC` behaviour. But this is just my point of view :-)

roadster31:
The traditional hook mechanics are still available. A module may use the shortcuts for rendering some pages and CSS, and a specific set of classes to implement sophisticated features.

Best of both worlds :)

roadster31:
@bibich , for CSS and JS, is it possible to pass filters ?

bcabarrou:
For what it's worth I would also be in favour of keeping things simple, i.e. class methods for all hooks. I can understand the appeal of syntaxic sugar though, and Thelia is not only made for me :)

:+1: for the hook cleaning commands in any case !

bibich:
@roadster31 it's not yet possible to add filters for css or js but it's conceivable. We just  have to find the right syntax.

nicolasleon:
:thumbsup: As long as both methods are available.

PR #1717, opened 2015-10-07
check stock for virtual product

aamant:
Fix the control of stock for vituels products before payment


roadster31:
Manually merged to master

PR #1722, opened 2015-10-08
Replaced parameter "locale" with "lang" in generated URL

bibich:
The `locale` parameter in generated URLs of products, categories, ... is ambiguous with the `lang` parameter. And changing the locale in URL does not change the lang of the displayed content.

As the pages are indexed with the parameter 'locale' and not 'lang' by search engines there is potentialy a problem. In fact, it only happens when the rewriting is disabled on your website, which is rather uncommon.

I 'm not sure that there won't be side effects with this PR. Will you please look it over ?

### DIFF
--- a/core/lib/Thelia/Core/Stack/ParamInitMiddleware.php
+++ b/core/lib/Thelia/Core/Stack/ParamInitMiddleware.php
@@ -134,10 +134,16 @@ class ParamInitMiddleware implements HttpKernelInterface
     protected function detectLang(TheliaRequest $request)
     {
         // first priority => lang parameter present in request (get or post)
-        if ($request->query->has("lang")) {
-            // The lang parameter may contains a lang code (fr, en, ru) for Thelia < 2.2,
-            // or a locale (fr_FR, en_US, etc.) for Thelia > 2.2.beta1
-            $requestedLangCodeOrLocale = $request->query->get("lang");
+        $requestedLangCodeOrLocale = $request->query->get("lang");
+
+        // add a fallback on locale parameter
+        if (null === $requestedLangCodeOrLocale) {
+            $requestedLangCodeOrLocale = $request->query->get("locale");
+        }
+
+        // The lang parameter may contains a lang code (fr, en, ru) for Thelia < 2.2,
+        // or a locale (fr_FR, en_US, etc.) for Thelia > 2.2.beta1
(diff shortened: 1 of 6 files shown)


core/lib/Thelia/Core/Stack/ParamInitMiddleware.php
```diff
@@ -134,7 +134,14 @@ protected function defineCurrency(TheliaRequest $request)
     protected function detectLang(TheliaRequest $request)
     {
         // first priority => lang parameter present in request (get or post)
-        if ($request->query->has("lang")) {
+        $requestedLangCodeOrLocale = $request->query->has("lang");
+
+        // add a fallback on locale parameter
+        if (null === $requestedLangCodeOrLocale) {
```
lunika: I think you have to compare with `false`. The `has` method will return `true` or `false`

core/lib/Thelia/Core/Stack/ParamInitMiddleware.php
```diff
@@ -134,7 +134,14 @@ protected function defineCurrency(TheliaRequest $request)
...
+            $requestedLangCodeOrLocale = $request->query->has("locale");
+        }
+
+        if (null !== $requestedLangCodeOrLocale) {
```
lunika: same here

core/lib/Thelia/Core/Stack/ParamInitMiddleware.php
lunika: Or don't use the `has` but the `get` method.

core/lib/Thelia/Core/Stack/ParamInitMiddleware.php
```diff
@@ -134,7 +134,14 @@ protected function defineCurrency(TheliaRequest $request)
...
             // The lang parameter may contains a lang code (fr, en, ru) for Thelia < 2.2,
             // or a locale (fr_FR, en_US, etc.) for Thelia > 2.2.beta1
             $requestedLangCodeOrLocale = $request->query->get("lang");
```
lunika: You have to do something here, this line is wrong and depends on what you will do with the lines above.

bibich: Yes, you're right, this is nonsense :)

gillesbourgeat:
:+1:

PR #1729, opened 2015-10-13
fix database path in Thelia Update

Alban-io:
THELIA_CONF_DIR constant already contains DIRECTORY_SEPARATOR.


gillesbourgeat:
:+1:

PR #1735, opened 2015-10-15
fix export date range

Alban-io:
Currently, The order export range is a hard coded value in the [export-modal.html](https://github.com/Alban-io/thelia/commit/06e31020167bec70679e6e2ced9361d830e805d4#diff-96a64801063438570e597deaa43e952cL33) template.
This PR adds a [new ExportHandler property](https://github.com/Alban-io/thelia/commit/06e31020167bec70679e6e2ced9361d830e805d4#diff-1ecbc36bc80f92908003ab7ed8882012R44) which determine if an export is a "range export".


Alban-io:
Include in #1815

PR #1739, opened 2015-10-19
Module interface, add the missing method update

gillesbourgeat:


core/lib/Thelia/Module/BaseModuleInterface.php
```diff
@@ -26,5 +26,7 @@ public function preDeactivation(ConnectionInterface $con = null);
 
     public function postDeactivation(ConnectionInterface $con = null);
 
+    public function update($currentVersion, $newVersion, ConnectionInterface $con = null);
```
lovenunu: What is the goal of that method ?
And what are ```$currentVersion``` and ```$newVersion``` ? Are they Thelia or Module version ? :)

gillesbourgeat: This method already exists, she is just missing in the interface BaseModuleInterface.

https://github.com/gillesbourgeat/thelia/blob/8ccd16d01e68aca05cd0cd9c1180a642006ceb0b/core/lib/Thelia/Module/BaseModule.php#L517-L527

lovenunu: :+1:

roadster31:
By adding the default "null" to $con parameter, you break the compatibility with exiting modules which uses the signature without the default 'null' value, which was valid until now.

core/lib/Thelia/Module/BaseModule.php
```diff
@@ -521,7 +521,7 @@ public function install(ConnectionInterface $con = null)
      * @param string $newVersion the new module version, as defined in the module.xml file
      * @param ConnectionInterface $con
      */
-    public function update($currentVersion, $newVersion, ConnectionInterface $con)
+    public function update($currentVersion, $newVersion, ConnectionInterface $con = null)
```
roadster31: By adding the default "null" to $con parameter, you're breaking the compatibility with exiting modules which uses the signature without the default 'null' value, which was valid until now.

Please remove this 'null' default.

roadster31: By adding the default "null" to $con parameter, you're breaking the compatibility with exiting modules which uses the signature without the default 'null' value, which was valid until now.

Please remove this 'null' default.

core/lib/Thelia/Module/BaseModuleInterface.php
roadster31: Please remove the 'null' default to preserve BC.

PR #1747, opened 2015-10-23
Fixed success_url check for contact form

roadster31:
because `$this->retrieveSuccessUrl()` never return null...


gillesbourgeat:
Thanks !

PR #1750, opened 2015-10-27
Add EQUAL to product loop filter by min or max

lopes-vincent:
In product loop :

Change min_stock Criteria::GREATER_THAN to GREATER_EQUAL

Change min_weight Criteria::GREATER_THAN to GREATER_EQUAL
Change max_weight Criteria::LESS_THAN to LESS_EQUAL

Because if we set min_stock to 1 we expect to get the products with at least 1 product in stock not 2.


gillesbourgeat:
Thanks !

PR #1751, opened 2015-10-27
Fix for undefined currency exchange rate

roadster31:
When an exchange rate could not be found the error page was displayed, and the following currencies exchange rate were not updated.

Now, the problem is reported to the user on the currencies page, and the exchange rate update process terminates normally.

img-2015-10-27 09 49 09

The PR also contains a small improvement in the BaseAdminController::errroPage() method, which now logs a the error in the Thelia log.


gillesbourgeat:
:+1: Thanks !

PR #1752, opened 2015-10-27
add addValues method in EnumListType

Alban-io:


Alban-io:
Because I don't want to exclude Iterator Object ;)

core/lib/Thelia/Type/EnumListType.php
```diff
@@ -36,6 +36,17 @@ public function addValue($value)
         }
     }
 
+    /**
+     * @param array $values
+     * @since 2.3.0
+     */
+    public function addValues($values = array())
```
lunika: IMO if you don't want to use type hinting here, you have to test if `$values` is an `array` or an instance of `Traversable` and throw an exception if it's not the case

core/lib/Thelia/Type/EnumListType.php
```diff
@@ -36,6 +36,17 @@ public function addValue($value)
         }
     }
 
+    /**
+     * @param array $values
```
lunika: This is not really true with your implementation

Alban-io:
I agree with your opinion @lunika, i will fix it.

gillesbourgeat:
Thanks !

PR #1754, opened 2015-10-27
add homepage redirection in showLoginAction

Alban-io:
add homepage redirection in showLoginAction ( /admin/login ) if admin is already authenticate.


gillesbourgeat:
Thanks !

PR #1758, opened 2015-10-28
Image loop improvement

InformatiqueProg:
Add width/height for image and original image

### DIFF
--- a/core/lib/Thelia/Core/Template/Loop/Image.php
+++ b/core/lib/Thelia/Core/Template/Loop/Image.php
@@ -361,11 +361,19 @@ class Image extends BaseI18nLoop implements PropelSearchLoopInterface
                 // Dispatch image processing event
                 $this->dispatcher->dispatch(TheliaEvents::IMAGE_PROCESS, $event);
 
+                $originalImageSize = getimagesize($sourceFilePath);
+
+                $imageSize = getimagesize($event->getCacheFilepath());
+
                 $loopResultRow
                     ->set("IMAGE_URL", $event->getFileUrl())
                     ->set("ORIGINAL_IMAGE_URL", $event->getOriginalFileUrl())
                     ->set("IMAGE_PATH", $event->getCacheFilepath())
                     ->set("PROCESSING_ERROR", false)
+                    ->set("IMAGE_WIDH", $imageSize[0])
+                    ->set("IMAGE_HEIGHT", $imageSize[1])
+                    ->set("ORIGINAL_IMAGE_WIDH", $originalImageSize[0])
+                    ->set("ORIGINAL_IMAGE_HEIGHT", $originalImageSize[1])
                 ;
             } catch (\Exception $ex) {
                 // Ignore the result and log an error
@@ -377,6 +385,10 @@ class Image extends BaseI18nLoop implements PropelSearchLoopInterface
                         ->set("ORIGINAL_IMAGE_URL", '')
                         ->set("IMAGE_PATH", '')
                         ->set("PROCESSING_ERROR", true)
+                        ->set("IMAGE_WIDH", '')
+                        ->set("IMAGE_HEIGHT", '')
+                        ->set("ORIGINAL_IMAGE_WIDH", '')
+                        ->set("ORIGINAL_IMAGE_HEIGHT", '')
                     ;
                 } else {
                     $addRow = false;
--- a/core/lib/Thelia/Core/Template/Loop/Image.php
+++ b/core/lib/Thelia/Core/Template/Loop/Image.php
@@ -108,7 +108,8 @@ class Image extends BaseI18nLoop implements PropelSearchLoopInterface
             Argument::createBooleanTypeArgument('force_return', true),
             Argument::createBooleanTypeArgument('ignore_processing_errors', true),
             Argument::createAnyTypeArgument('query_namespace', 'Thelia\\Model'),
-            Argument::createBooleanTypeArgument('allow_zoom', false)
+            Argument::createBooleanTypeArgument('allow_zoom', false),
+            Argument::createBooleanTypeArgument('with_image_size', false)
         );
 
         // Add possible image sources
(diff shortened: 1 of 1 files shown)


gillesbourgeat:
Hello,

You break four unit tests
https://travis-ci.org/thelia/thelia/builds/87851013

Best regards

gillesbourgeat:
It is not possible to apply your pull request in Thelia.
This change makes the image loop too greedy in disk resource.

thank you for your proposition.

InformatiqueProg:
And if i add an argument  'with_image_size' ?

gillesbourgeat:
No,
It is possible for you, to delegate this behavior in a module.

```php
class MyLoopImage extends Image
{
    /**
     * Use this method in order to add fields in sub-classes
     * @param LoopResultRow $loopResultRow
     * @param object|array $item
     */
    protected function addOutputFields(LoopResultRow $loopResultRow, $item)
    {
        // your code
    }
}
```

lunika:
> This change makes the image loop too greedy in disk resource.

Can you explain why @gillesbourgeat  ?

InformatiqueProg:
Okay,
i will do it in a module.
I see some post who says that getimagesize load entire image, that's why ?

roadster31:
getimagesize() is well known to be slow on remote files, but as far as I know, there's no specific performance issues on local files. Thus, adding another getimagesize() will have little impact on the loop performance.

Let's make some tests :)

roadster31:
Here are the results for 50 iterations of the following code on a live server, with and without getimagesize(), for 1572 images :

```
{loop type="product" name="testprod"}
    {loop type="image" name="test" product=$ID}
    {/loop}
{/loop}
```
The results are showing an average cost of 0.46 seconds when the getimagesize() version is used. A very small difference.

Giving this results, I'm supporting this PR, as getting image sizes may be useful for the front designer.

img-2015-10-29 10 49 10

core/lib/Thelia/Core/Template/Loop/Image.php
```diff
@@ -361,11 +361,19 @@ public function parseResults(LoopResult $loopResult)
                 // Dispatch image processing event
                 $this->dispatcher->dispatch(TheliaEvents::IMAGE_PROCESS, $event);
 
+                $originalImageSize = getimagesize($sourceFilePath);
```
bibich: I think you should use `$event->getOriginalFileUrl()` instead of `$sourceFilePath` here.

InformatiqueProg:
Okay,
i'll change this, and test it.
While i am, what about add an argument 'with_image_size' to prevent this ?

core/lib/Thelia/Core/Template/Loop/Image.php
InformatiqueProg: You're sure ?
 $event->getOriginalFileUrl() return an url => http://mydomains.fr/cache/images/product/on_air_2-6.jpg
it seem that remote file are not the best choice for getImageSize()...

Soldras:
Hello,
the IMAGE_WIDTH and IMAGE_HEIGHT parameters are useless because if you don't pass width or height, you get the original file size. 

I don't know what you want to do with this parameters but the loop give you ORIGINAL_IMAGE_URL and IMAGE_URL. In front office if you need image size, it's the parameters you have pass to the loop.
If you need orignal size to perform some front processing, you can do it with some js.

It's not a need for Thelia so it's better to do a module.

roadster31:
> the IMAGE_WIDTH and IMAGE_HEIGHT parameters are useless because if you don't pass width or height, you get the original file size. 
No. You can give only width or only height.

Additionally, without IMAGE_WIDTH and IMAGE_HEIGHT, you have to *copy* the numeric width and height passed to the loop, which may cause errors / inconsistencies.

> If you need orignal size to perform some front processing, you can do it with some js.
It's very easy to provide the dimension parameters from the loop. Why bother the front developer with complicated stuff to get them ?

lunika:
:+1: with @roadster31 :-)

Soldras:
> No. You can give only width or only height.
It's true but if you have one size element you can calculate the other.

> It's very easy to provide the dimension parameters from the loop. Why bother the front developer with complicated stuff to get them ?
It's better to let front dev to front developers. Image size is like the spoon, it doesn't exit in front dev (Cf Matrix ).
If you have a basic image you rules it by css. If you want some dynamic view you do it in js.

roadster31:
Once again, providing these dimensions have not any performance impact, and may help some front developers. So let's provide them !

I'll not argue any longer. I suggest a vote to decide if we'll accept this PR.

Soldras:
The real question is : 
it's interesting to loose 6% answer time to 0.1% use case ?

lunika:
> The real question is :
it's interesting to loose 6% answer time to 0.1% use case ?

This is why loop were designed for. Be the more generic possible and provide much information as possible.

roadster31:
6% ? The average impact is 0.46 second for 1572 image, which gives an average impact of 0.00003 second per image.

Soldras:
> This is why loops were designed for. Be the more generic possible and provide much information as possible.

Being generic is important. But being generic not involve being slow because we want to cover all particular use case. This type of use case must be cover by module.

> 6% ? The average impact is 0.46 second for 1572 image, which gives an average impact of 0.00003 second per image. No big deal !

8.057 x 100 / 8,5156 = 94,6145... So theres is 5,4% performance lost. It's valid for each image.

roadster31:
> 8.057 x 100 / 8,5156 = 94,6145... So theres is 5,4% performance lost. It's valid for each image.

True. 5,4% for a single image. This means 0.00003 seconds. Is it a performance problem ?

roadster31:
Well, stop feeding the troll.

Let's vote.

lunika:
Let's vote : 

OpenStudio : :-1: 
Community : :+1: 

Hoops, stop feeding the troll :-)

bilhackmac:
It can be a performance problem if images are hosted on a CDN (incoming feature)

Alban-io:
@bilhackmac :+1:

InformatiqueProg:
Nobody want's my argument 'with_image_size' ? :disappointed_relieved:
It will only use on somes loop...

Soldras:
@InformatiqueProg :
Dalek

lunika:
@Soldras he already explain twice, please read the discussion !

This kind of argument already exist and is used in the `Product` loop for example : https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Core/Template/Loop/Product.php#L357

Soldras:
@lunika : he didn't explain why he need these outputs

InformatiqueProg:
Like @roadster31 said, 
We need width and height for front end integration of some image (JS gallery, etc.)

lunika:
I apologize for my troll...

Have fun guys

PR #1760, opened 2015-10-29
Set order status as paid in FreeOrder::pay()

roadster31:
When the The FreeOrder module is used to "pay" an order, the order remains in the "not paid" status.

This PR fixes the problem, and sets the status as "paid" in FreeOrder::pay().


gillesbourgeat:
:+1: Thanks

PR #1764, opened 2015-11-03
Missing class

InformatiqueProg:
Toolbar pagination bottom is hidden if not


gillesbourgeat:
Hello,

there is the same problem in the templates : sale.html, brand.html, cart.html

InformatiqueProg:
The "row" class is missing for ```sale.html``` and ```brand.html```, but not for ```cart.html```...

gillesbourgeat:
:+1: indeed for car.html

PR #1769, opened 2015-11-08
Increase API key size to 48

jmaupetit:
An even length for hexadecimal strings increases compatibility with other languages (like python) that systematically refuse to convert odd hexadecimal strings to binary data. This could be blocking if someone tries to consume thelia's API.


lunika:
:+1: This PR doesn't break BC So it will be great to merge it in version 2.1 and 2.2 too.

gillesbourgeat:
Thanks !

Issue #1655, opened 2015-08-26
PSE ref don't change if we change product ref


InformatiqueProg:
In admin:
- Create a product (without attributes)
- Save it
- Change the ref
- Save it
-Go to front, product 's page, you will see old ref

The PSE 's ref has not been changed

PS: Thelia 2.1.5

Mertiozys:
Because the product has no attributes, there are no PSEs in the "Price" tab of a product edit page, and so there is no way to edit PSE's ref which is displayed in the front-office. That's why editing the product's ref doesn't change the displayed PSE's ref in FO.
4 solutions :
- set the good ref at the beginning and never change it :-p
- in the template, check how many PSEs the product has. If there's only one, display the product's ref instead of the PSE's one.
- in the back-office, add a field in the "Price" tab to allow ref modification (or remove the special display if there is only 1 PSE to set the table listing PSEs as when there are many of them)
- modify product update event in order to set the same ref to its PSE when the product's one is changed

InformatiqueProg:
Your first solution should be the better... but we are in the real life
For me, the fourth solution must to be done, i 'll try to do this.
Thanks

PR #1790, opened 2015-11-23
Update the default PSE ref when the product ref is updated

bibich:
Close #1656 
related to issue #1655


gillesbourgeat:
Thanks !

PR #1801, opened 2015-11-26
Fixed cart duplication conditions at user login/logout

roadster31:
This PR is a fix for #1798 : the cart should be duplicated when a user with a permanent discount is logging in.

The code for deciding if the current cart should be duplicated has been factorized and improved.


roadster31:
This PR is finisted.

Issue #1774, opened 2015-11-13
Image file name is automatically set as image title


roadster31:
The image file name (eg. some-image.png) is automatically set as the image title. In most cases, the image file name is not meaningful. As this file name is returned by the $TITLE image loop variable, one's cannot distinguish a "real" title from a "file name" title.

Leaving the title empty is much more interesting, as one's can detect this in the front template, and replace it with something meaningful, something like `{$TITLE|default:$someMeaningfulTitle}`

PR #1802, opened 2015-11-26
The image file name is no longer the default image title

roadster31:
This is a fix for  #1774. An image did not require a default title (documents do).


gillesbourgeat:
Thanks !

PR #1805, opened 2015-11-26
Add locale param for module_config smarty

lopes-vincent:
- if locale specified get the config value of this locale instead of session locale
- if no locale specified still get value for session locale


gillesbourgeat:
Thanks !

PR #1807, opened 2015-11-27
Fix: TheliaFormValidator parsable translation

Asenar:
The first param inside translation->trans() was splited in 2 string
    making the parser not working for that particular case.
I could have modified the regex expression in core/lib/Thelia/Action/Translation.php
but the result was incertain.


Asenar:
rebased on master again :)

gillesbourgeat:
Thanks !

PR #1810, opened 2015-11-30
Add output variables total price for orderProduct loop

InformatiqueProg:
And use it for fix current wrong total value in the templates


gillesbourgeat:
Thanks !

PR #1817, opened 2015-12-01
tinymce plugin fullscreen fix

InformatiqueProg:
Resolve bug: navbar is over tinymce in fullscreen mode

### DIFF
--- a/local/modules/Tinymce/templates/backOffice/default/tinymce_init.tpl
+++ b/local/modules/Tinymce/templates/backOffice/default/tinymce_init.tpl
@@ -41,6 +41,15 @@
             "fullscreen code youtube importcss"
         ],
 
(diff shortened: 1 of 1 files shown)


Soldras:
:+1:

gillesbourgeat:
Hello,
:-1: 
The solution is in css, not in js.
The css class ```.navbar-static-top``` has a bad zIndex.

InformatiqueProg:
I'm agree with you, a CSS correction is better solution (z-index must be less than 100)

InformatiqueProg:
or if need to keep z-index: 1000
body.mce-fullscreen .navbar-static-top {
    display: none;
}

gillesbourgeat:
No, you just have to remove the z-index.
(Must be tested)

Issue #1575, opened 2015-07-16
White page after having deleted a product from the cart


Vyndi:
After adding one or several  products to the cart, if a customer follows the path : 
1.	« See the cart »
2.	« Proceed with the purchase »
3.	Previous 
And then tries to remove an item from the cart by clicking on « delete », he is redirected to a blank page

Thelia version 2.1.2

gillesbourgeat:
The problem is that you used the security token

PR #1822, opened 2015-12-02
Proposition : Remove the force refresh for the links with token

### LINKED ISSUE #1811: security / invalid token /cart/delete/34?_token=46 front uses backOffice/error.html
TokenAuthenticationException in TokenProvider.php line 97: Tried to validate an invalid token

How to reproduce:

- add some items to your cart
- delete one of them (you are redirected to the cart page)
- refresh the page 

/cart/delete/34?_token=4644334205063df86a0260d4b12fe641some_invalid_token

This show an error page (templated from backOffice/default/error.html ) and show a link to /admin .

gillesbourgeat:
This is a proposition.
Another solution is to store a unique token per link, but it's expensive in resource.

Resolve #1811 #1575

### DIFF
--- a/core/lib/Thelia/Tools/TokenProvider.php
+++ b/core/lib/Thelia/Tools/TokenProvider.php
@@ -50,7 +50,7 @@ class TokenProvider
     protected $tokenName;
 
     /**
-     * @param SessionInterface    $session
+     * @param Request $request
      * @param TranslatorInterface $translator
      * @param $tokenName
      */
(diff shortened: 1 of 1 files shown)


roadster31:
The "one token per link" seems a good solution.

What do you mean saying 'expensive in resource' ? Generating a few tokens (e.g., less than 100.000) takes a few microseconds, and the generated tokens are stored in the user session.

IMO, the cost of "one token per link" is not significant.

PR #1823, opened 2015-12-02
Add states/provinces concept

bibich:
:construction: Work in progress :construction:

The objective of this PR is to separate states/provinces of countries. 
For now, the concept of states/provinces was managed in country model which was not the best way.

With this PR country can have states/provinces managed in a new model `State`. A lot of related stuff should be modified :

- addresses / order_addresses
- tax rules
- shipping zone
- delivery module interface
- a lot of templates to modify
- configuration page to migrate from previous system, because update can't be done easily and securely.   
- ...

In addition, I added some improvements :

- country visible or not (state too)
- better zip code management for country
- revamp tax rules edit page :scream:

This is a big PR ( sorry for the review @gillesbourgeat :) ) and there is still some stuff to do, tests to add, ... But if you see anything wrong, tell me !


roadster31:
The schema.xml was changed, but the MySQL Workbench model was not updated.

Could you please update it ?

bibich:
Yes. I'll do it. The documentation too.

roadster31:
Some translations are missing in the B.O. Exemple :

img-2015-12-31 11 27 38

roadster31:
The tax rule "configuration" (may be renaming it "Applicable Countries ans States" would be better...) is not selected when the tab is clicked, making the form difficult to understand.

img-2015-12-31 11 30 44

bibich:
For now, translation keys and english translations are present.  
I will update [Crowdin repo](https://crowdin.com/project/thelia). The french and other languages will be translated on Crowndin and then synced in github.

roadster31:
Well, to be honest, I don't understand at all how "configurations" are working.

bibich:
The way it works didn't change. But it's not easy to explain.
The UI has just changed a little.

roadster31:
I did not succeed to create a 5.5% tax rule for France Métropolitaine.

PR #1835, opened 2015-12-09
Missing combinations product in PDF delivery

InformatiqueProg:
- center quantity like PDF invoice
- change cols width


gillesbourgeat:
Thanks !

PR #1840, opened 2015-12-11
Use routId in url function of UrlGenerator.

Doud63:
add two parameter : routeId and router
default router is router.front or router.admin

### DIFF
--- a/local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
+++ b/local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
@@ -12,24 +12,33 @@
 
 namespace TheliaSmarty\Template\Plugins;
 
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Routing\Router;
 use TheliaSmarty\Template\SmartyParser;
 use TheliaSmarty\Template\SmartyPluginDescriptor;
 use TheliaSmarty\Template\AbstractSmartyPlugin;
 use Thelia\Tools\TokenProvider;
 use Thelia\Tools\URL;
 use Thelia\Core\HttpFoundation\Request;
-use Thelia\Core\Translation\Translator;
 
 class UrlGenerator extends AbstractSmartyPlugin
 {
     protected $request;
 
     protected $tokenProvider;
+    /** @var ContainerInterface */
+    private $container;
 
-    public function __construct(Request $request, TokenProvider $tokenProvider)
+    /**
+     * @param Request            $request
+     * @param TokenProvider      $tokenProvider
+     * @param ContainerInterface $container         Needed to get all router.
+     */
+    public function __construct(Request $request, TokenProvider $tokenProvider, ContainerInterface $container)
     {
         $this->request = $request;
         $this->tokenProvider = $tokenProvider;
+        $this->container = $container;
     }
 
     /**
@@ -44,7 +53,18 @@ class UrlGenerator extends AbstractSmartyPlugin
         // the path to process
         $current = $this->getParam($params, 'current', false);
         $path  = $this->getParam($params, 'path', null);
-        $file  = $this->getParam($params, 'file', null); // Do not invoke index.php in URL (get a static file in web space
+        // Do not invoke index.php in URL (get a static file in web space
+        $file = $this->getParam($params, 'file', null);
+        $routeId = $this->getParam($params, 'route_id', null);
+        // select default router
+        if ($this->request->fromAdmin()) {
+            $defaultRouter = 'admin';
+        } elseif ($this->request->fromFront()) {
+            $defaultRouter = 'front';
+        } else {
+            $defaultRouter = null;
+        }
+        $routerId = $this->getParam($params, 'router', $defaultRouter);
 
         if ($current) {
             $path = $this->request->getPathInfo();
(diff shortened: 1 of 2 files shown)


bcabarrou:
I would suggest using `route` instead of `routeId`. This is more consistent with other parameters in my opinion (`router` is also an id for example).

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,42 @@ public function generateUrlFunction($params, &$smarty)
             );
         }
 
+        if ($id != null && $router != null) {
```
bcabarrou: Should be `!==` tests ?

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,42 @@ public function generateUrlFunction($params, &$smarty)
...
-        $excludeParams = $this->resolvePath($params, $path, $smarty);
+            if ($file !== null) {
+                $path = $file;
+                $mode = URL::PATH_TO_FILE;
+            } elseif ($path !== null) {
+                $mode = URL::WITH_INDEX_PAGE;
+            } else {
+                throw new \InvalidArgumentException(
+                    Translator::getInstance()->trans(
+                        "Please specify either 'path' or 'file' parameter in {url} function."
```
bcabarrou: Now can also specify a route id, this message should probably be updated accordingly.

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,42 @@ public function generateUrlFunction($params, &$smarty)
             );
         }
 
+        if ($id && $router) {
+            // get url by router and id
+            /** @var Router $router */
+            $router = $this->container->get($router);
+            if (!$router) {
+                throw new \InvalidArgumentException(sprintf("Router %S doesn't exist", $router));
```
roadster31: Use %s with a lowecase "s"

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -44,7 +54,18 @@ public function generateUrlFunction($params, &$smarty)
         // the path to process
         $current = $this->getParam($params, 'current', false);
         $path  = $this->getParam($params, 'path', null);
-        $file  = $this->getParam($params, 'file', null); // Do not invoke index.php in URL (get a static file in web space
+        // Do not invoke index.php in URL (get a static file in web space
+        $file = $this->getParam($params, 'file', null);
+        $id = $this->getParam($params, 'routeId', null);
```
roadster31: $routeId would be more convenient, $id is a too generic name

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,42 @@ public function generateUrlFunction($params, &$smarty)
             );
         }
 
+        if ($id && $router) {
+            // get url by router and id
```
roadster31: if (null === $id && null === $router)

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,42 @@ public function generateUrlFunction($params, &$smarty)
...
+            /** @var Router $router */
+            $router = $this->container->get($router);
+            if (!$router) {
```
lunika: This will never happen, if the service does not exists, an exception will be thrown (and I think its a `ServiceNotFoundException` )

Doud63: Exact, thanks

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
Doud63: Shift !! thanks

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
Doud63: Nop
when $router or $id = ''
if (null !== $id) return true
and 
if ($id) return false

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,39 @@ public function generateUrlFunction($params, &$smarty)
             );
         }
 
+        if ($routeId && $router) {
+            // get url by router and id
+            /** @var Router $router */
+            $router = $this->container->get($router);
```
MaximeMorille: I think it would be better to have `$router = $this->container->get($routerId);`

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -44,7 +54,18 @@ public function generateUrlFunction($params, &$smarty)
         // the path to process
         $current = $this->getParam($params, 'current', false);
         $path  = $this->getParam($params, 'path', null);
-        $file  = $this->getParam($params, 'file', null); // Do not invoke index.php in URL (get a static file in web space
+        // Do not invoke index.php in URL (get a static file in web space
+        $file = $this->getParam($params, 'file', null);
+        $routeId = $this->getParam($params, 'routeId', null);
```
gillesbourgeat: Hello,
Please, use SnakeCase for the argument name.

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,39 @@ public function generateUrlFunction($params, &$smarty)
...
+            $router = $this->container->get($routerId);
 
+            $excludeParams = $this->resolvePath($params, $path, $smarty);
 
-        if ($file !== null) {
-            $path = $file;
-            $mode = URL::PATH_TO_FILE;
-        } elseif ($path !== null) {
-            $mode = URL::WITH_INDEX_PAGE;
+            $url = $router->generate(
```
gillesbourgeat: Please return the absolute url. 
You must add the argument ```Router::ABSOLUTE_URL```

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
```diff
@@ -57,25 +78,39 @@ public function generateUrlFunction($params, &$smarty)
             );
         }
 
+        if ($routeId && $routerId) {
+            // get url by router and id
+            /** @var Router $router */
+            $router = $this->container->get($routerId);
 
+            $excludeParams = $this->resolvePath($params, $path, $smarty);
```
gillesbourgeat: Are you sure this line is necessary ?

local/modules/TheliaSmarty/Template/Plugins/UrlGenerator.php
gillesbourgeat: @bcabarrou :+1:

gillesbourgeat:
Thanks !

PR #1847, opened 2015-12-15
[WIP] Fix : return to default language on the use of "prev" and "next"

vigourouxjulien:
Fix : return to default language on the use of "prev" and "next" arrows in the product, category, content, folder and brand page.


bibich:
Your PR could't be merged. There is a problem with the insert.sql, it looks like you have deleted all translations. Could you please run this command in the root dir of your Thelia :

```bash
php Thelia generate:sql --locales='de_DE,en_US,es_ES,fr_FR'
```

Thanks

bibich:
Sorry, but why do you restore the default language. If this admin has changed the language it's better to use this one ?

gillesbourgeat:
@bibich :+1: 

It's necessary to add the new config variable in the update file ```2.3.0-alpha1.sql.tpl```
Example : https://github.com/thelia/thelia/blob/master/setup/update/sql/2.2.0-alpha1.sql#L32

lopes-vincent:
Outdated PR

PR #1851, opened 2015-12-17
tinymce update

ThomasArnaud:
Add the possibility to choose in which text areas the editor will be used.
Add the summary and the conclusion to the available text areas.


gillesbourgeat:
Thanks !

PR #1852, opened 2015-12-17
[proposition] Add the possibility to disable the generation of url for the loops

gillesbourgeat:
The loops returns an url although there are no need.
This PR adds argument ```return_url```  that can be used for disable the generation of urls.
The default value for argument ```return_url``` is ```true```


roadster31:
What is the goal of this PR ? Is there a perfomance problem with URL generation ?

gillesbourgeat:
Not with the generator, but with too many useless request.

Example for a site that has 300 brands, 200 categories.
The product edit page will make 500 useless request on rewriting table.

The product edit page will make 500 useless request on rewriting table.

If the rewrite table contains many url, the product edit page becomes very slow.

Other example for the FrontOffice with faceted search engine.

Other example for the FrontOffice with faceted search engine.

PR #1853, opened 2015-12-18
Add condition match for cart item include quantity

### LINKED ISSUE #1804: Condition code promo
il serait bien dans la création d'un code promo d'avoir une condition sur le nombre d'items dans le panier et non pas seulement pour le nombre de produits.

Alban-io:
Close #1804


gillesbourgeat:
Thanks !

PR #1855, opened 2015-12-18
Add hook "home.dashboard" and module HookDashboard for display the dashboard

vigourouxjulien:
Insert a new hook "home.dashboard" and add a new module HookDashboard for display the dashboard in administration's home page.


gillesbourgeat:
Resolve by #2045

PR #1857, opened 2015-12-21
correction of hookblack : order.tab

Yochima:
There is a wrong use of capture order_tab_tab in order-edit.html (probably do to a copy/paste from folder-edit.html)


gillesbourgeat:
Thanks !

PR #1861, opened 2015-12-22
Add links to the appropriate pages

ThomasArnaud:
Add links to the appropriate pages on the count of the customers, products, orders etc at the top of the page.


bibich:
I think you should first test if the current admin has right to view this pages :

```smarty
{loop type="auth" name="can_view" role="ADMIN" resource="admin.customer" access="VIEW"}
...
{/loop}
```

PR #1862, opened 2015-12-22
add a next and previous button to the order edit page

ThomasArnaud:


bibich:
This PR will be included in a new one

PR #1864, opened 2015-12-23
add next/prev button for customer edit page

ThomasArnaud:


bibich:
This PR will be included in a new one

Issue #1863, opened 2015-12-23
Colissimo shipping zones has changed


jodeq:
When I update my shipping price with the last collisimo pricing I see that shipping zones from colissimo has changed.
Here the pdf : http://www.colissimo.fr/particuliers/envoyer_un_colis/decouvrir_loffre_colissimo/Tarifs_colissimo/att00002941/PDF_Tarifs2015_AELGP_DepartFranceMetropolitaine.pdf

Would you like if I provide for you the new shipping zones in thelia with the right prices ?

Regards.

bibich:
Thanks. Yes, it would be great to provide us the new configuration :+1: 
You can do it in a new Pull Request. More info on how to proceed here : http://doc.thelia.net/en/documentation/contribute.html#pull-request

Thanks

jodeq:
PR is merged.
This issue could be closed

bibich:
thanks

PR #1865, opened 2015-12-23
Mise à jour des tarifs colissimo

jodeq:
Colissimo price update


jodeq:
PR pour #1863

bibich:
Thanks. 
Could you also include the translations please. They should be defined in the `setup/I18n` directory.  
It's recommended to use the translation tool in the backOffice > Configuration > Translation. Item to translate : **Thelia installation** in **developper mode**   
You can only do it for en_US and fr_FR.

bibich:
Thanks

PR #1868, opened 2015-12-24
Allow to open order-edit.html template with a specific module tab

Yochima:
Actually, the order-edit tab could be selected (with param "tab") but the tab-pane was never active


gillesbourgeat:
Thanks !

PR #1872, opened 2015-12-24
Add next/prev button

ThomasArnaud:
Add next/prev buttons for orders and customers.
Modify the loops of brands, categories, folders and contents so that the queries to get the next and previous objects are sent only when it is needed.


gillesbourgeat:
Hello,
Please complete the documentation for the order loop. http://doc.thelia.net/en/documentation/loop/order.html

gillesbourgeat:
Thanks !

PR #1875, opened 2015-12-29
Update coupon-update.html

SimonWaters:
The user can currently break the admin UI by entering mark-up in a coupon name. 

This is reported as XSS in my security testing, but I don't think it is a security issues as there are only administrators with full privileges (who could steal each others passwords, if so motivated, by other means).

I am not experience SMARTY user, so alternative solutions to same may be preferable.


roadster31:
Thanks for reprting this. The problme has been solves by removing the quotes around {$couponCode}

`{intl l='Editing coupon "%title"' title={$couponCode}}`

PR #1876, opened 2015-12-29
Update form.html

SimonWaters:
If a user includes HTML mark-up ("<",">") then the display of the coupon name in the page title is double escaped.

I added nofilter, and the title is escaped once, I haven't security tested other inclusions of the "form.html" template with this change.

I'm not experienced SMARTY user, other fixes to this issue may be preferable.


roadster31:
A fix for this problem is included from 2.3.4.

Thanks for reporting, and for the fix :)

PR #1877, opened 2015-12-29
Update shipping-configuration-edit.html

SimonWaters:
Users are currently able to break the UI by putting HTML mark-up in a shipping zone name.


roadster31:
A fix for this problem is included from version 2.3.4.

Thanks :)

PR #1881, opened 2015-12-29
Display only the zones affected to Colissimo in the bo

ThomasArnaud:


ThomasArnaud:
The class extends the order loop and there was a bug with the with_prev_next_info argument.

Issue #1728, opened 2015-10-12
Wrong annotations in order model


lopes-vincent:
In Model/Order.php all choosenAddress (delivery and invoice) getters and setters have wrong annotations, for example :

```
    /**
     * @return Address
     */
    public function getChoosenDeliveryAddress()
    {
        return $this->choosenDeliveryAddress;
    }
```

the type is set to address but in reality choosenDeliveryAddress is address id not the object.

manuelbcd:
Solved. Waiting to pull request revision. Regards

manuelbcd:
Sorry, it was solved previously.  https://github.com/thelia/thelia/pull/1701

bibich:
Fixed by #1701

PR #1886, opened 2016-01-01
Wrong annotations in Order model #1728

manuelbcd:
Corrected types and information in annotations as requested in issue.


roadster31:
These fixes are included in #1701

manuelbcd:
Ok sorry. Thanks for your advice. Regards

PR #1891, opened 2016-01-05
Add missing primary key

Alban-io:
Add missing primary key in coupon_customer_count and ignored_module_hook tables.


gillesbourgeat:
Thanks !

PR #1894, opened 2016-01-06
Set content loop default order as manual

Mertiozys:


bibich:
Instead of changing the default order of this loop (which could generate side effects), it would be better to update the documentation.

PR #1901, opened 2016-01-07
Colissimo export update

Mertiozys:
Add link to order and to customer
Add package weight


ThomasArnaud:
Hello.
You should do the same on the master.

PR #1905, opened 2016-01-08
Display products that belongs to category even if it's not there default category

bilhackmac:


gillesbourgeat:
Hello,
This pull request generates the problem indicated on this issue https://github.com/thelia/thelia/issues/1941

PR #1906, opened 2016-01-08
Fix coupon create form data

Alban-io:
Prevent conflict with [Update Coupon Action](https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Controller/Admin/CouponController.php#L205)


gillesbourgeat:
:+1:

PR #1912, opened 2016-01-13
Fix for feature loop with product filter

zorn-v:
There filter only by feature template related to product. It fetch all features in that template, not only related to requested product.
And because this, it VERY SLOW (for every feature there is addition query to retrieve feature value) while opening product if many features in one template.


zorn-v:
Up

roadster31:
Thanks

PR #1914, opened 2016-01-14
The module list in the translation page is now ordered by module code instead of module title

roadster31:
As each item starts with the module code, it's easier to find a module if the list is ordered by module code.


bibich:
Thanks

PR #1920, opened 2016-01-17
Fixed coupons conditions label translation

roadster31:
This PR fixes the label of coupons conditions, where the english (en_US) title for products, categories and countries was always displayed, whatever the current B.O language is.


gillesbourgeat:
Thanks !

PR #1922, opened 2016-01-18
Move the prices from a json to a config

ThomasArnaud:
The prices and the freeshipping boolean are now saved in the table module_config.


roadster31:
There is a still a consistency problem between colissimo config and areas.

As the config is not stored in a table with a FK to the area table, if an area is deleted, the change is not automatically cascaded to Colissimo config, and there's no listener to handle the change in Colissimo.

ThomasArnaud:
Thanks for your comment, I added a listener to handle the deletion of an area in the config.

roadster31:
:+1:

bibich:
Thanks Céline

Issue #1927, opened 2016-01-20
Make hook development easier by throwing exceptions in dev mode


roadster31:
To make hook development easier, the Thelia/Core/DependencyInjection/Compiler/RegisterHookListenersPass should throw exceptions in development mode instead of silently logging an alert in the Thelia log

PR #1928, opened 2016-01-20
Hook DI alert messages thrown as exceptions in dev. mode

roadster31:
This PR is a fix for issue #1927. In development mode, the alert messages are also thrown as exceptions, making typos and logic errors much more visible.

This will give an immediate answer to the very common interrogation "why this fscking hook does not show up ?!!", and saves YEARS of accumulated headaches ;)


gillesbourgeat:
Thanks ! This is also a good idea for many head ;)

PR #1939, opened 2016-01-25
Add visible and visible_reverse values in Product Loop order argument

Alban-io:


roadster31:
Don't forget to update the documentation.

PR #1940, opened 2016-01-25
autofill $version array for update process

### LINKED ISSUE #1934: Can't update Thelia 2.2.1 with Thelia v2.2.2 (web update)
il me dit :
"Il semblerait que Thelia soit déjà installé sur ce serveur. Merci de vérifier la configuration et d'apporter les corrections nécessaires avant de ré-essayer. "

Le problème vient du fichier ``` thelia/core/lib/Thelia/Install/Update.php ```
La version 2.2.1 est 2 fois et pas de de 2.2.2
        '39' => '2.2.1',
        '40' => '2.2.1',

Asenar:
Before each release, you have to edit a static property of
the Update class, and this is pretty complex when this has to match
several branch version (for example 2.0.x/2.1.x/2.2.x).

This modification remove the static and use the content of SQL_DIR
to fill the array Update->version (non static) with all version.

With that, your release process will not require anymore to update that
file.

Last point: if you use the update process from the master branch, you
will update to a «not released yet version», which can still change
later), but as it's a development branch, so this should not be a problem.
Add to that this provide an easy way to test the update process before
a release and without having to edit anything.

Close #1934


Asenar:
I just replaced the function `strnatcmp()` by `version_compare()`, which works better to set the right version order.

I also made a comparison between the version property before and after the changes, you can see here:
latest-screenshot

on the left, the `print_r(self::$version)` at the end of the constructor without my PR
on the right, the `print_r($this->version)` at the end of the constructor with my changes

You will notice a few version number are missing on the right, because there isn't any sql to execute. The update process still works fine :)

Notice also the current development version is shown. This is a positive side effect, because now we can easily test the upgrade process (and otherwise, we may have trouble using "new files with not updated database").

Asenar:
updated !
I added a second commit to automate the version numbering, but this time it's related to the install/release process.

The purpose is to have only one place to set the version number :)

core/lib/Thelia/Install/Update.php
```diff
@@ -207,14 +167,29 @@ public function process()
...
+            $updateConfigVersion = [
+                [':name' => 'thelia_version'        , ':value' => $currentVersion['version'],],
+                [':name' => 'thelia_major_version'  , ':value' => $currentVersion['major'],  ],
+                [':name' => 'thelia_minus_version'  , ':value' => $currentVersion['minus'],  ],
+                [':name' => 'thelia_release_version', ':value' => $currentVersion['release'],],
+                [':name' => 'thelia_extr_version'   , ':value' => $currentVersion['extra'],  ],
+            ];
+
+            foreach ($updateConfigVersion as $args) {
+                $sql = 'UPDATE `config` SET `value`=":value" WHERE `name`=":name";';
```
gillesbourgeat: Hello,
Why not use `ConfigQuery::write()`.

Example : 
```php
ConfigQuery::write('thelia_version', $currentVersion['version']);
```

Asenar: :+1: done

gillesbourgeat:
Thanks !

PR #1942, opened 2016-01-25
BackOffice CSS change for DateTimePicker disabled

InformatiqueProg:


gillesbourgeat:
Closed, because identical to #1943

InformatiqueProg:
Heu...
Yes and No:
a) It's for Thélia 2.2, #1943 for master
b) some less file have changed in master, the CSS file generated are not the same

PR #1944, opened 2016-01-27
check current stable release

Asenar:


Asenar:
Should I change the backOffice/home template to use a local request related to `Thelia/Install/Update->getWebVersion()` ?

Asenar:
I rebased what I've done over #1965 then rebased all on current master :)

That PR now uses `Version::parse()` to secure the response of version.php

gillesbourgeat:
Informations for my commits.

Fixes an undesirable behavior of the pull request #1788

This pull request deletes all Adminincludes unseless in the core modules, and also adds the ability to display messages from the update file.

After the update 2.3.0-alpha1, composer does not deleted the "Adminincludes". So, the module interfaces were duplicated

gillesbourgeat:
Thanks !

PR #1946, opened 2016-01-27
Fix: TaxType class only if extension == php

Asenar:
This fix a problem when php.ini files (or any non php file) are present
in core/lib/Thelia/TaxEngine/TaxType/
see http://thelia.net/forum/viewtopic.php?id=11849


gillesbourgeat:
Thanks !

PR #1949, opened 2016-01-28
fix loop folder output not correct content_count

zzuutt:
normally the loop folder return the variable $CONTENT_COUNT ->  the number of visible contents for this folder  
currently $COUNTENT_COUNT return the number of all contents, not just visible

This fix corrects this bug

### DIFF
--- a/core/lib/Thelia/Core/Template/Loop/Folder.php
+++ b/core/lib/Thelia/Core/Template/Loop/Folder.php
@@ -81,7 +81,8 @@ class Folder extends BaseI18nLoop implements PropelSearchLoopInterface, SearchLo
                 'manual'
             ),
             Argument::createIntListTypeArgument('exclude'),
-            Argument::createBooleanTypeArgument('with_prev_next_info', false)
+            Argument::createBooleanTypeArgument('with_prev_next_info', false),
+            Argument::createBooleanOrBothTypeArgument('content_count_visible', 1)
         );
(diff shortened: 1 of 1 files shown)


bibich:
Firstly, it could be *dangerous* as you change the previous behavior, setting `visible` to `true`.
Secondly, you never call the function with a parameter, and don't use the visible argument of the loop.
Thirdly, why not apply this to `CHILD_COUNT` and `countChild` function.

Maybe using the visible argument of the loop for the 2 functions could be a solution ?

zzuutt:
Bonjour, la boucle `function `normalement doit retourner `$CONTENT_COUNT` qui correspond, d'après la doc, au nombre de contenu visible du dossier passé en argument
Or elle retourne le nombre total des contenus visible et non visible.

ce fix rétablit le fonctionnement correspondant à la doc

zzuutt:
I added parameter `content_count_visible` in loop folder
CONTENT_COUNT return
`content_count_visible  = true` -> only the content visible
`content_count_visible = false` -> only the content not visible
`content_count_visible = '*'` -> both (all contents)

by default return only the content visible (to match the description of the doc [http://doc.thelia.net/en/documentation/loop/folder.html](url))

roadster31:
Your changes have been manually merged to the master branch, and will be available in version 2.3.4.

Thank you :)

PR #1952, opened 2016-01-29
Add missing join condition operator in customer loop

bcabarrou:
The previous form happens to work (because the value is a boolean), but reverse the intended condition since the value argument will never be used.

The `ModelCriteria::addJoinCondition` is meant to be used with an explicit `?` value placeholder as far as I can tell : this call does not work on the current Propel master version.


gillesbourgeat:
Thanks !

PR #1965, opened 2016-02-02
Update process 2.3.0-alpha2 : Deletes the "Adminincludes" useless in the core modules

gillesbourgeat:
Requires PR #1940

Fixes an undesirable behavior of the pull request #1788

This pull request deletes all `Adminincludes` unseless in the core modules, and also adds the ability to display messages from the update file.

After the update 2.3.0-alpha1, composer does not deleted the "Adminincludes". So, the module interfaces were duplicated


gillesbourgeat:
This pull request is included in #1944

PR #1970, opened 2016-02-04
CartDuplicationEvent to provide both original and duplicated cart to listeners

roadster31:
This PR introduces the CartDuplicationEvent, which stores the original cart so that listeners will get both duplicated and original carts.

Currently, only the duplicated cart is passed, and the listeners have no simple way to get the original one.

CartDuplicationEvent extends CartEvent, thus preserving BC.


bibich:
Thanks

PR #1907, opened 2016-01-08
Administrator email management and features

roadster31:
This PR is a fix for #1510. It adds an address email to the administrator profile. This address email can now be used to login just like the login name.

An administrator could now recover a lost password, just like a regular customer.

Minor improvement: add a language selector on back-office login page.

PR #1971, opened 2016-02-04
Administrator email management and features

### LINKED ISSUE #1510: There's no way for and administrator to retrieve a lost password
Thus, a negligent administrator wiil be locked out the back-office !

A way to solve this problem would be to store the administrator email address in its profile data, and implement the same password reset process as in the front office.

roadster31:
This PR is a fix for #1510. It adds an address email to the administrator profile. This address email can now be used to login just like the login name.

An administrator could now recover a lost password, just like a regular customer.

Minor improvement: add a language selector on back-office login page.

Close #1907 
Resolve #1510


gillesbourgeat:
Thanks !

roadster31:
:)

PR #1974, opened 2016-02-05
Fix: SEO - richSnippet (price, itemCondition "new" + breadcrumb)

Asenar:


templates/frontOffice/default/product.html
```diff
@@ -153,12 +153,15 @@ <h1 class="name"><span itemprop="name">{$TITLE}</span><span id="pse-name" class=
                             <meta itemprop="category" content="{$TITLE}">
                         {/loop}
                         {* List of condition : NewCondition, DamagedCondition, UsedCondition, RefurbishedCondition *}
-                        <meta itemprop="itemCondition" itemscope itemtype="http://schema.org/NewCondition">
+                        {if $IS_NEW}
```
bibich: I don't think it's the same new. `newCondition` indicates that this product has not been used. The `IS_NEW` of Thelia indicates that this product is new in the catalog.

Asenar: indeed … I will check what can match the "is new" thing

Asenar:
I left what was in the template, just fixed the syntax to validate google tools.

Do you want I totally remove the `NewCondition` related code ? Or should I put it in a system variable ?

bibich:
thanks

PR #1976, opened 2016-02-05
all Thelia files are EOF unix style + fix email admin

Asenar:
end.php is now unix style instead of dos


bibich:
in fact all these files are in dos EOL : 

- web/install/connection.php
- web/install/end.php
- web/install/config.php
- web/install/bdd.php
- web/install/permission.php
- web/install/header.php
- web/install/bootstrap.php

could you convert them too ?

Asenar:
I converted end.php because I edited it, but sure :)

bibich:
Oops I havn't seen your real modification (line 64) as all lines was in red in the diff.
sorry :)

gillesbourgeat:
Thanks !

PR #1980, opened 2016-02-08
Fix module version field length

bilhackmac:
Update schema to increase module version field to 25 chars.
With only 10 chars update function may be call on each `module:refresh`, `cache:clear` commands.

Image your module is at version `1.0.0.beta.2`, only `1.0.0.beta` was stored.
Image your module is at version `1.0.0.alpha`, only `1.0.0.alph` was stored.

In both case update function was call for these modules with truncated version as `$currentVersion` and current (real) version as `$newVersion`.

For information `BaseModule::update` signature is
```php
public function update($currentVersion, $newVersion, ConnectionInterface $con = null)
```


bibich:
Thanks

PR #1985, opened 2016-02-09
Delivery payment events

bibich:
The goal of this PR is to give more control on how a delivery or a payment module can be used for an order.

The EventDispatcher component is used to make things possible.

## For Delivery modules

An event is now dispatched to get 2 information :

- is module valid for an address/cart ?
- the postage amount to deliver the parcel to an address

It no longer only uses the delivery country to determine the validity and the amount, but an address and a cart. 

Other modules can listen this event too and control more precisely :

- the valid shipping areas for a module
- the amount, tax rules of the shipment
- the estimated delivery date

There is a backward compatibility, so if a module doesn't use the event system, Thelia uses `isValidDelivery` and `getPostage` functions of the main modules class. 

## For Payment modules

An event is now dispatched to get 2 information :

- is module valid for an address/cart ?
- does module manage the stock on order creation

There is a backward compatibility, so if a module doesn't use the event system, Thelia uses `isValidPayment` and `manageStockOnCreation` functions of the main modules class. 

## Others changes

- updated `BaseModuleInterface`
- in `order-delivery-module-list.html` :
    - added some HTML attributes to select easily HTML tag related to a module
    - postage amount is hidden when equal to 0
    - added an estimated delivery date
- updated `cart_postage` smarty function

## Example

```php
/**
 * Class Order
 * @author Julien Chanséaume <julien@thelia.net>
 */
class Order implements EventSubscriberInterface
{
    /**
     * Call for every delivery modules
     */
    public function preDeliveryGetPostage(DeliveryPostageEvent $event)
    {
        if ($event->getModule()->getCode() == 'StorkDelivery') {
            $event->setValidModule(false);
            $event->stopPropagation();
        }
    }

    /**
     * Call for every delivery modules
     */
    public function deliveryGetPostage(DeliveryPostageEvent $event)
    {
        Tlog::getInstance()->debug('[TheliaDebug] deliveryGetPostage');
    }

    /**
     * Call for every delivery modules
     */
    public function postDeliveryGetPostage(DeliveryPostageEvent $event)
    {
        if ($event->isValidModule()) {
            if (null === $deliveryDate = $event->getDeliveryDate()) {
                $deliveryDate = new \DateTime("@".strtotime("today + 5 days"));
            }

            if (0 !== intval(ConfigQuery::read("kernel_panic", 0))) {
                $deliveryDate->add(new DateInterval('P5D'));
            }

            $event->setDeliveryDate($deliveryDate);
        }
    }

    /**
     * Call only for colissimo module
     * values are hard coded here for the example
     */
    public function deliveryGetPostageColissimo(DeliveryPostageEvent $event)
    {
        /*
        $module = $event->getModule();
        $address = $event->getAddress();
        $cart = $event->getCart();
        $country = $event->getCountry();
        $state = $event->getState();
        */

        $event->setValidModule(true);
        $event->setPostage(5.00);
        $deliveryDate = new \DateTime("@".strtotime("today + 1 week"));
        $event->setDeliveryDate($deliveryDate);

        // we can add additional data that we be added to the delivery loop
        // if we want to display it, you have to change the template
        $event->addAdditionalData('TITLE', "Colissimo plus");
        $event->addAdditionalData('EXTRA', "lorem ipsum ...");

        // stop the dispatched of the event
        $event->stopPropagation();
    }

    /**
     * @inheritdoc
     */
    public static function getSubscribedEvents()
    {
        return [
            TheliaEvents::MODULE_DELIVERY_GET_POSTAGE => [
                ['postDeliveryGetPostage', 64],
                ['deliveryGetPostage', 128],
                ['preDeliveryGetPostage', 256],
            ],
            TheliaEvents::getModuleEvent(TheliaEvents::MODULE_DELIVERY_GET_POSTAGE, 'Colissimo') => ['deliveryGetPostageColissimo', 128],
        ];
    }
}
```


lopes-vincent:
:+1:

InformatiqueProg:
:star2:

roadster31:
Please don't forget to update the documentation, as this is a useful feature for module authors.

bibich:
yes, I planned to write documentation. But it's still a work in progress, and I would like to hear your opinion first.

gillesbourgeat:
Thanks !

PR #1992, opened 2016-02-11
Add method to find category path

bilhackmac:


gillesbourgeat:
:+1:

Can you add [tag @since](http://www.phpdoc.org/docs/latest/references/phpdoc/tags/since.html) and squash your commits

Thanks !

bilhackmac:
Done :+1:

bilhackmac:
Wait before merging… there is a mistake !

bibich:
thanks

PR #1995, opened 2016-02-11
Added order search options, improved search page

roadster31:
This PR adds the order invoice reference, delivery number and transaction reference to the searched items in the back-office.

It also improves the search page, by displaying a message instead of an empty table.

img-2016-02-11_18-11-32


gillesbourgeat:
Thanks !

PR #2001, opened 2016-02-16
Check PHP version before trying to do anything in install process

roadster31:
(same player shoot again)

### DIFF
--- a/web/install/header.php
+++ b/web/install/header.php
@@ -128,4 +128,3 @@ if ($context == "install") { ?>
                     </div>
 <?php
 }
-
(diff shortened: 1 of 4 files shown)


bcbrr:
Isn't it also done in https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Install/CheckPermission.php#L120 (called here: https://github.com/thelia/thelia/blob/master/web/install/header.php#L44) ? It may not belong here though.

roadster31:
No. Old PHP fails on namespace declaration in header.php :

``Parse error: syntax error, unexpected T_STRING, expecting T_CONSTANT_ENCAPSED_STRING or '(' in /home/www/*************/web/thelia/web/install/header.php on line 21`

PR #2005, opened 2016-02-19
Use a wider version requirement on thelia/installer for setup/

### LINKED ISSUE #1979: Composer warning "dependencies will break in the future" : `composer-plugin-api:1.0.0`
Today, a `composer install` display the following message : 

```bash
$ composer install -o
The "thelia/installer" plugin requires composer-plugin-api 1.0.0, this *WILL* break in the future and it should be fixed ASAP (require ^1.0 for example).
[...]
```

The versions of these dependencies should be updated.

bcbrr:
The exact version requirement can cause compatibility issues.

Fixes #1979.

### DIFF
--- a/setup/composer.json
+++ b/setup/composer.json
@@ -2,7 +2,7 @@
     "name": "thelia/setup",
     "type": "thelia-local",
     "require": {
-        "thelia/installer": "1.1",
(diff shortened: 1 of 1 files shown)


gillesbourgeat:
Thanks !

PR #2015, opened 2016-02-26
navbar title allready hidden on "SM size" while navbar-toggle is visible

InformatiqueProg:
capture
PS: `visible-xs` is depreced as of bootstrap v3.2.0
http://getbootstrap.com/css/#responsive-utilities-classes


gillesbourgeat:
Thanks !

PR #2017, opened 2016-03-01
Fix issue 1090 hook code unicity

### LINKED ISSUE #1090: Hook code unicity is not checked when creating a new hook
When a new hook is created, Thelia doesn't check code unicity. 

For example, it's possible to create several 'home.top' hooks.

We can check unicity in the `HookCreationForm`, but we have to define a class validator, as we need both hook code and hook type fields value. Is it possible in a Thelia form ?

TomPradat:
Function to check code unicity during HookCreationForm added with a Callback
Resolve #1090

### DIFF
--- a/core/lib/Thelia/Form/HookCreationForm.php
+++ b/core/lib/Thelia/Form/HookCreationForm.php
@@ -12,9 +12,13 @@
 
 namespace Thelia\Form;
 
+use Propel\Runtime\ActiveQuery\Criteria;
+use Symfony\Component\Validator\Constraints\Callback;
 use Symfony\Component\Validator\Constraints\NotBlank;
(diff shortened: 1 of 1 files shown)


core/lib/Thelia/Form/HookCreationForm.php
```diff
@@ -81,6 +87,19 @@ protected function buildForm()
         ;
     }
 
+    public function checkCodeUnicity($code, ExecutionContextInterface $context){
+        $count = HookQuery::create()->filterByCode($code)->count();
```
bibich: You have to check the type of the hook too, as you can have two hooks with the same code, one used in the front office, the other in the back office.

PR #2022, opened 2016-03-02
Fix: saveDescription on module update process

Asenar:


gillesbourgeat:
:+1:

gillesbourgeat:
Hello,
This proposition create a problem when a user have change manually the title or description in the backOffice.
After an update of module, the title or description changed by user will be crushed

PR #2026, opened 2016-03-04
Fix: add preg_quote for preg_replace base URL

Asenar:


gillesbourgeat:
Hello,
Do you have an example of code to demonstrate the problem encountered ?

Asenar:
I don't have, I got an error when I first configured Thelia, and this changes fixed by problem (but maybe the real problem was elsewhere).

An error can happen when an argument name contains a `/` , `.*` or other regex thing.

    $url = URL::getInstance()->absoluteUrl('path/to/action', array("cat1/cat2" => "v1", "p3" => "v2"));`

roadster31:
Manually merged to master

PR #2028, opened 2016-03-05
Fixed translation regexp prefix for templates

roadster31:
The current regexp catches `l="invoice.imprint"` in `{elsehook rel="invoice.imprint"}`. The new one don't.


gillesbourgeat:
:+1:  

A small example for the pdf invoice.html : https://regex101.com/r/uU4pP6/3

PR #2036, opened 2016-03-12
Take THELIA constants into account in CheckPermission

radmiraal:
This change changes the CheckPermission class so it uses the
THELIA constants defined in the bootstrap.

I took the liberty to remove the usage of class constants
as they do not seem to be referenced anywhere else in the code.

### DIFF
--- a/core/lib/Thelia/Install/CheckPermission.php
+++ b/core/lib/Thelia/Install/CheckPermission.php
@@ -26,22 +26,9 @@ use Symfony\Component\Translation\Translator;
  */
 class CheckPermission extends BaseInstall
 {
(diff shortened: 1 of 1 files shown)


roadster31:
> I took the liberty to remove the usage of class constants as they do not seem to be referenced anywhere else in the code.
Unfortunately, some modules may use them.

radmiraal:
agree about the BC if that would be the case, but imho that would be a bad practice as they should use the constants in the bootstrap (at least that's my understanding after using the stuff for 5 minutes :p )

radmiraal:
another way would be to add 2 constants to the bootstrap for the media and session dir...

PR #2042, opened 2016-03-14
Avoid having too many results in the backoffice search page

gillesbourgeat:
This pull request fix to 25 the number of search results by type to avoid too long loading times or of memory overflow.

### DIFF
--- a/templates/backOffice/default/search.html
+++ b/templates/backOffice/default/search.html
@@ -8,6 +8,18 @@
 {block name="main-content"}
 {$searchTerm = {$smarty.get.search_term|trim}}
 
+{capture name="more_results"}
+    <tr>
+        <td colspan="99">
+            <div class="alert alert-info text-center">
+                {intl l="More results are available."}
+                <br/>
+                {intl l="Please refine your search to see the other results."}
+            </div>
+        </td>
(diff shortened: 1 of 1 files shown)


roadster31:
Is it possible to add a notice when more than 25 results are available ?

gillesbourgeat:
@roadster31 This requires making additional queries that are very greedy in this page.
Why not, an under search that allowing of paginate the results of a type if the number of results returned by the loop is equals 25.
This under search is accessed by a link under the search of the first list.
This feature will be for Thelia 2.4.

roadster31:
@gillesbourgeat to achieve this without additional queries, limit the loop to 26 instead of 25, hide the 26th result with a Smarty test, and check if ($LOOP_COUNT > 25).

gillesbourgeat:
@roadster31 ok, but for now there will be just a message and not a link allowing paginate the search.

roadster31:
Sure. This is just a workaround to the lack of pagination, to warn the user that only the first 25 results are displayed.

gillesbourgeat:
@roadster31 Done ! This suits your ?

roadster31:
Wonderful :heart:

PR #2047, opened 2016-03-16
Template variables are now available in hook methods

roadster31:
This PR makes the variables of the template from which the hook is called available to Hook methods, using two new methods of `BaseHookRenderEvent` : 
- `getTemplateVariable()` to get the value of a template variable,
- `hasTemplateVariable()` to check if a variable is defined in the template from which the hook is called


roadster31:
TODO : inject variables in the hook to make them available to quick hooks (http://doc.thelia.net/en/documentation/modules/hooks/hook_create.html#improvements-since-thelia-23)

PR #2053, opened 2016-03-18
forgot Table colissimo_freeshipping

zzuutt:


gillesbourgeat:
Hello,
This table does not exist in the schema of the Colissimo mdodule.
Why would you add this table ?

zzuutt:
colissimo module use this table
there is no schema in the config folder in 2.3

Asenar:
This table was in local/modules/Colissimo/Config/schema.xml in 2.2.2

in 2.3 @ThomasArnaud removed schema.xml in 028c18d

PR #2055, opened 2016-03-18
Version::parse() Allow -dev version + allow 2 digit only test (for module version)

Asenar:
This commit makes the method `Version::parse()` working if Thelia version has a `-dev` suffix.

This way, install and update process can be tested more easily if you set version number like `2.3.0-alpha2-dev`. For example:

1) I install Thelia from the repository:master branch (which will install version `2.3.0-alpha2`.
2) I add or remove some fields/column, propel rebuild
3) I can test the update process from `2.3.0-alpha2` to `2.3.0-alpha2-dev`

This PR also add a 2nd parameter `releaseMandatory` (default to true). If set to false, it allows version number with only 2 digits, if this method is used for module version or other things.

### DIFF
--- a/core/lib/Thelia/Tools/Version/Version.php
+++ b/core/lib/Thelia/Tools/Version/Version.php
@@ -111,8 +111,11 @@ class Version
             (?<major>[0-9]+)\.
             (?<minus>[0-9]+)\.
             (?<release>[0-9]+)
-            -?(?<extra>[a-zA-Z0-9]*) # extra_version will also match empty string
-        )$`x";
+            -?(?<extra>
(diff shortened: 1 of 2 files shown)


bcbrr:
I'm not clear on what `releaseMandatory` does. Is it just so that version numbers such as `2.3` can be used ? If so, where are they considered to be relative to `2.3.0` for example (before, after, the same) ?

Asenar:
if the `releaseMandatory` is set to false, the `Version::parse` method can be used to test module numbers, which are sometimes on 2 numbers only. I didn't choose well the push title …

PR #2057, opened 2016-03-19
Front template minor syntax fixes

roadster31:
This PR fixes the front template syntax which confuses the PHPStorm parser. This is mostly due to useless quotes, such as `{loop ... param="{$somevar}" ... }`, which could be expressed as `{loop ... param={$somevar} ... }`

We're keeping the `{$somevar}` form instead of using `$somevar`, to be sure the `somevar` variable content is escaped


gillesbourgeat:
Thanks !

Issue #2059, opened 2016-03-21
TaxRule action introduces a compatibility break


roadster31:
Before 2.3.0-alpha1, TaxRuleEvent::getCountryList() is expected to return an array of countries.

From 2.3.0-alpha1, TaxRuleEvent::getCountryList() is expected to return an array of arrays of countries and states IDs, possibly JSON encoded.

Modules such as ImportT1 which uses the TaxRuleEvent are now broken.

gillesbourgeat:
Resolved by #2060
Thanks !

PR #2060, opened 2016-03-21
Fix for #2059 - Action\TaxRule.php BC break

roadster31:
This PR introduces a method in Action\TaxRule.php to ensure a compatibility with the 2.2.2 country arrays.
In 2.2.2, the TaxRueEvent::getXXXCountryList() methods returned an array of country IDs. [ country ID, country ID ...].
From 2.3.0-alpha1, these functions are expected to return an array of arrays, each one containing a country ID and a state ID. [ [ country ID, state ID], [ country ID, state ID], ...].

The method checks the $obj parameter, and create a 2.3.0-alpha1 compatible return value if $obj is expressed using the 2.2.2 form.

Some sanity checks have also been added.


gillesbourgeat:
Thanks !

PR #2062, opened 2016-03-23
Remove composer dependency leafo/lessphp

gillesbourgeat:


roadster31:
You have to remove also the Assetic\Filter\LessphpFilter class, and the less_legacy.assetic.filter service, which depends on leafo/lessphp

Issue #2061, opened 2016-03-23
Deleting a product with a free text feature value breaks the feature behaviour.


roadster31:
When a product is deleted, if it has a free text feature, the related Feature Availability is not deleted, and it causes the Feature to be changed from a free text feature to a list feature.

To reproduce the problem:

1) create a new feature, and do not add any feature availability

![img-2016-03-23_20-12-54](https://cloud.githubusercontent.com/assets/2197734/13997716/afeb0ab2-f133-11e5-9064-74d4e27f2ee1.png)

2) in a product, give a value to this feature

![img-2016-03-23_20-14-14](https://cloud.githubusercontent.com/assets/2197734/13997745/d88dffb0-f133-11e5-87f5-6f0f4692c022.png)

everything is still fine on the feature side :

![img-2016-03-23_20-14-47](https://cloud.githubusercontent.com/assets/2197734/13997770/f1aa7ca8-f133-11e5-8553-6454192bf8c7.png)

3) delete the product

4) go back to see the feature definition : you have now one feature availability, which is the free text value of the deleted product.

![img-2016-03-23_20-16-03](https://cloud.githubusercontent.com/assets/2197734/13997815/27c5c270-f134-11e5-8240-e9a3c2ff1929.png)

Deleting the feature av. along with the product in Action\Product::delete() may solve the problem. But deleting a product using a SQL query will still cause the problem, compromising the database integrity.

Changing the database to set the free_text_value flag in the feature_av table instead of the feature_product table) will solve the problem, even if the product is deleted using a SQL command. A useless line will be left in the feature_av table, though.

gillesbourgeat:
Nice catch ;)

PR #2063, opened 2016-03-23
Fix for #2061, minor CS improvements

roadster31:
This is a simple workaround for #2061.

The problem pointed by #2061 still occur if a product is deleted with something like ProductQuery::create()->findPk(some_product_id)->delete();

Moving the free_text_value in the feature_av table (and reworking the related code) would be a much better solution.


gillesbourgeat:
Thanks !

PR #2066, opened 2016-03-25
Fix the problem of position if a product or content in several sections and folders

### LINKED ISSUE #132: Product and Content variable ranking depending on the category / folder
We should be able to provide different product/content order depending on the category/folder.

Currently product/content are only ranked depending on their main category/folder.

### LINKED ISSUE #1941: ability to sort product in each category
Currently, the product position is determined by the field `product.position` and concern only its main category. it would be more flexible if a product could be positioned differently in each category, so merchant can use a different default order in each categories, putting products forward according to their needs (the default theme does not show "default sorting" but it can be easily added)

Current workaround (without coding) is to create the main category containing all products, and then adding others category as "secondary". But this make some mess in the backoffice product lists.

gillesbourgeat:
Fix #1941 
Fix #132
Close #1905


Asenar:
Thanks a lot !!

bibich:
Thanks

PR #2078, opened 2016-03-30
Fixed checkbox and radio automatic rendrering

roadster31:
The "checked" status of checkboxes and radios was not correctly managed by form-field-attributes-renderer.html

This PR fixes the problem, and removes an extra "true" string that was floating around.


gillesbourgeat:
Hello,
Can you squash your commits !
Thanks !

roadster31:
Done !

gillesbourgeat:
Thanks !

PR #2080, opened 2016-04-01
Core loops, no function addoutputfields

TomPradat:
In some loops the function addoutputfields didn't exist which prevented to extend from this loop.
So i added missing addoutputfields functions where needed.


gillesbourgeat:
Thanks !

PR #2082, opened 2016-04-04
WIP fix issue #2003 : product and pse ref in invoice template

### LINKED ISSUE #2003: BO: PSE ref in order page & invoice template
Actually on order-edit BO template, the product ref, title and feature are listed, but not the PSE ref matching the feature. Sellers often need it to retrieve products, it should be added in the default BO template.
Also, product and PSE refs are absent from PDF invoice templates and should also be added.

TomPradat:
Product ref and pse ref added to invoice and delivery template and to emails.

tmp1

tmp2

tmp3

Resolve #2003
To do : update.php in 2.4

### DIFF
--- a/templates/email/default/order_confirmation.html
+++ b/templates/email/default/order_confirmation.html
@@ -76,7 +76,7 @@
             {/if}
                 <tr>
                     <td style="border-bottom:1px solid #000">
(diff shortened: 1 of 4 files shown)


gillesbourgeat:
Thanks !

PR #2089, opened 2016-04-06
fix #2087, check in db if current admin exists

### LINKED ISSUE #2087: the session is style active after an admin is deleted
I tested that issue on both 2.2.2 and master branch.

I created 2 admins: `user1` and `user2`, and logged in with both on them from 2 different browser (firefox and firefox, with 2 differents profiles)

logged with `user1`, I deleted `user2`
`user2` has still full access to the admin.

I also tried to remove all cache but the session seems only cookie dependant.

Asenar:
This commit checks if the current logged in admin (from session) still
exists in the database, and if the password has not changed.

### DIFF
--- a/core/lib/Thelia/Core/Security/SecurityContext.php
+++ b/core/lib/Thelia/Core/Security/SecurityContext.php
@@ -16,6 +16,7 @@ use Symfony\Component\HttpFoundation\RequestStack;
 use Thelia\Core\HttpFoundation\Session\Session;
 use Thelia\Core\Security\Resource\AdminResources;
 use Thelia\Core\Security\User\UserInterface;
+use Thelia\Core\Security\UserProvider\AdminUserProvider;
 use Thelia\Model\Customer;
 
 /**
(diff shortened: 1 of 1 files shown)


Asenar:
This fixes #2087

patrick-mcdougle:
With the symfony security component, the AbstractToken has a `hasUserChanged()` method that is supposed to take care of this. Does Thelia use/extend the AbstractToken?

Asenar:
Hi, I suppose it does, feel free to commit the fix, and mention the issue 2087 in the log message.

bibich:
It would be preferable to implement your checking in the `getAdminUser` method, maybe with a caching system.

Asenar:
The problem is precisely the current method which uses a cache somewhere (I don't know well symfony), with the consequence of a deleted admin can still do anything he wants (including creating an other admin) if his session is still active (and same problem for customers but my solution doesn't work :/ )

TomPradat:
Why not deleting the session of this admin when we delete the admin from the backoffice ? Would solve all problems and not add additionnal controls

roadster31:
Discussion (and fixes) continue in PR #2359

PR #2090, opened 2016-04-06
Fix GenerateRewrittenUrlEvent

Alban-io:
Add missing getters and setters in GenerateRewrittenUrlEvent.


gillesbourgeat:
Thanks !

PR #2093, opened 2016-04-08
Fix #1662 add of hooks in pdf email and account-order

### LINKED ISSUE #1662: Missing hook form cart item in PDF, mails and account-order.html
We should add in these documents a hook like the `order-edit.product-list` hook of the BO.

TomPradat:
fix #1662


gillesbourgeat:
Thanks !

PR #2097, opened 2016-04-08
Fixed and improved cancel order processing

roadster31:
The `order_id` parameter of the `Front\Controller\OrderController::orderFailed()` method was ignored, in favor of the `order_id` request parameter. This behavior is wrong as the current request may not contains an `order_id` parameter.

Additionally, in some cases, the payment module does not return an order ID when the order is canceled by the client. Now, this situation is smoothly processed, and the order-cancel page is displayed instead of an error page.


gillesbourgeat:
Hello,
Can you rebase this pull request.

gillesbourgeat:
Thanks !

Issue #2108, opened 2016-04-17
All module routers have the same priority (150) in the Thelia chained router


roadster31:
When a module's router (created from routing.xml) is added to the Thelia chained router, the priority is always 150. Thus, when the chained router sorts its router list, the routers order is not deterministic, and we have no way to know which route will be chosen if the same route is registered by more than one module.

Using the module position + 150 instead of 150 as the router priority will solve this problem

PR #2109, opened 2016-04-17
Module routers priority improvement (issue #2108)

roadster31:
This PR changes the way module routers priority is defined when RegisterRouterPass adds them to the Thelia chained router. Priority is now 150 + the router position instead of just 150.

Thus, the routers are now checked using the module's position, and the way a route is calculated is now deterministic.


gillesbourgeat:
Thanks !

PR #2128, opened 2016-04-26
Fix checkout issues

Mertiozys:
Fix address state check in delivery cost estimation
Fix login error due to symfony update


gillesbourgeat:
Thanks !

PR #2140, opened 2016-05-11
Added missing "calendar" PHP extension to docker config

roadster31:
The calendar extension is required to display the B.O. home graph.


gillesbourgeat:
Thanks !

PR #2145, opened 2016-05-13
Fix for taxes & tax rules description display in Taxes rules page.

roadster31:
In the Taxes rules page, the taxes and tax rules descriptions are displayed in tables. As these descriptions may contains HTML, they should not be escaped.

This PR contains a minor CSS fix to prevent extra bottom space in table cells which contains <p>


gillesbourgeat:
Thanks !

PR #2146, opened 2016-05-13
Fix search in i18n fields when backend_context=1, and search improvements

roadster31:
When `backend_context` loop parameter is 1, the `default_locale_i18n` pseudo-column is not defined. However, hard-coded search requests appear is some loops (product, category, etc.), causing an SQL error if `backend_context`=1 and `search_in` is not empty.

Additionally, this PR improves the loop search system, by providing the `StandardI18nFieldsSearchTrait`, which manages searches in the standard internationalized fields (title, chapo, description and postscriptum).

In the back-office, customer search is now in `sentence` (%term%) mode, to allow incomplete word matches, such as part of an email address.


gillesbourgeat:
Thanks !

PR #2149, opened 2016-05-17
Fixed status_id parameter access

roadster31:
As \Thelia\Controller\Admin\OrderController::updateStatus() could be by a GET request, the status_id request parameter should not be read from request body, which is set by POST requests only


gillesbourgeat:
Thanks !

PR #2157, opened 2016-05-20
Prevent an infinite loop in new product dialog

roadster31:
In the categories.html admin template, an infinite loop may occur if a module has modified the behavior of the category loop.

Minor style fixes.


gillesbourgeat:
Thanks !

PR #2160, opened 2016-05-23
Added missing home.block 'class' parameter

roadster31:
This parameter is used in the block code : 

`<div class="{if $class}{$class}{else}col-md-4{/if}">`


gillesbourgeat:
Thanks !

PR #2170, opened 2016-06-02
Fix: #2166 array to string conversion when php setup/update.php

Asenar:
This fixes `PHP Notice:  Array to string conversion in /home/michael/git/github/thelia/setup/update.php on line 96` when trying to update from command line with `php setup/update.php`


gillesbourgeat:
Oops, My bad. Thanks !

PR #2178, opened 2016-06-07
Add a coupon type to offer a product

TomPradat:


TomPradat:
For the moment i'll do this coupon type in a new module.

PR #2181, opened 2016-06-13
Fix CSV export cached file size

bilhackmac:
Depending platform, generated export file real size and cached size may differ


gillesbourgeat:
:+1:

PR #2188, opened 2016-06-16
A more effective way to solve issue #2061

roadster31:
Deleting free text feature av. of a product is done in  Model\Product::postDelete(), so that features av. are deleted even if Model\Product::delete() is directly called.

This is a better solution than #2063 for solving issue #2061


gillesbourgeat:
Thanks !

PR #2189, opened 2016-06-16
Huge performance improvement in feature-availability loop

roadster31:


bibich:
Thanks.
Although it would be better to store free text values in feature_product table.

roadster31:
Or in a specifi i18n tabler, as free text values should be available in several languages.

PR #2198, opened 2016-06-21
Cancel coupon usage on order cancel

roadster31:
When an order is canceled, the use of coupons attached to this order is also canceled. For short, coupon usage count is incremented by 1, and the related OrderStatus coupon is marked as canceled.

This PR syncs the schema.xml file with the WorkBench model, and adds minor CS and unit test improvements


bibich:
Thanks

PR #2199, opened 2016-06-23
Update country

zzuutt:
Update country
'Biélorussie' not correct. It's 'Belarus'
 isoalpha2 'PW' corresponds to the country 'Republic of Palau'


bibich:
You should use the intl function to translate these names in the tpl.
you should also modify the insert.sql.tpl.
and then generate the sql files with : `php Thelia generate:sql --locales='de_DE,en_US,es_ES,fr_FR'`

PR #2203, opened 2016-06-27
Check if $error exist, specific for submit type

Doud63:
SubmitType doesn't contain $fieldVars['errors'].
Check if this var exist to add in template.


gillesbourgeat:
Thanks !

PR #2204, opened 2016-06-28
Add isEmpty(), to check if $data is empty.

Doud63:
Data attribute is private.
Add method isEmpty() to check if $data is empty, to check if export is empty.


bibich:
Thanks

PR #2206, opened 2016-06-29
add reset array pointer if $data is an array.

Doud63:
Because, if getData() include foreach on $data, key($data) return false and export is empty.


bibich:
thanks

PR #2209, opened 2016-07-01
Feature/protected modules

vigourouxjulien:
Add 3 fields in module table :
 - core : for distribution module if it set to "1", you can't delete the module
 - secure : if it set to "1", you can't desactivate the module
 - visible : if it set to "1", you can't see the module in the modules management

Add 3 options in the "module:generate" command : 
 --core : set core to "1"
 --secure : set secire to "1"
 --visible : set visible to "1"


core/lib/Thelia/Command/Skeleton/Module/module.xml
```diff
@@ -38,4 +38,6 @@
     -->
     <thelia>2.2.0</thelia>
     <stability>other</stability>
+    <mandatory>0</mandatory>
```
roadster31: "required" is the correct term.

vigourouxjulien: "required" already used for module dependencies

roadster31: Then "essential" is fine.

Issue #2218, opened 2016-07-15
Coupon condition CartContainsProducts


TomPradat:
In this condition, the isMatching method isn't working well, if we want to check if a product is not in the cart we'll get a problem

bibich:
closed by #2219

PR #2219, opened 2016-07-15
Fix coupons issues

TomPradat:
* fix cart contains product, cart contains categories : #2218 
* fix ajax response on coupon select type


bibich:
Thanks

PR #2227, opened 2016-07-29
Fix for two problems with CART_FINDITEM event processing

roadster31:
1) The CartEvent dispatched with CART_FINDITEM was the one received by Thelia\Action\Cart::addItem(), thus stopping event propagation will stop the whole cart add process.

2) Thelia\Action\Cart::findCartItem() was not aware of a cart item set in the event by event listeners with a higher priority, thus this cart item is always overwritten.


bibich:
Thanks

PR #2236, opened 2016-08-16
fixes #2233: add missing isset in AddressCountryValidationTrait

### LINKED ISSUE #2233: Missing isset in AddressCountryValidationTrait::verifyZipCode and verifyState breaks customer profile update
Thelia version 2.3.2

CustomerController::updateAction uses form CustomerProfileUpdateForm, which inherits the AddressCountryValidationTrait from AddressCreateForm. But CustomerProfileUpdateForm removes the "country" parameter in its buildForm method, causing AddressCountryValidationTrait methods to fail because of a missing index.

NicolaF:
fixes #2233: add missing isset in AddressCountryValidationTrait


bibich:
Good catch, but I think that it would be better to remove the `state` field from the buildForm method `CustomerProfileUpdateForm` class :

```
$this->formBuilder
    ...
    ->remove('country')
    ->remove('state')
    ...
```

PR #2241, opened 2016-08-17
Last URL view for catalog

InformatiqueProg:
save last URL view for catalog (category/product) and get it with `{navigate to="catalog_last"}`


roadster31:
Thanks. I've updated the documentation

InformatiqueProg:
Thanks for the documentation (i forget to do it)

PR #2258, opened 2016-09-08
Implemented new DeliveryModuleExInterface

roadster31:
This PR introduces the DeliveryModuleExInterface, which adds the $address parameter to the getPostage() and isValidDelivery() methods, so that module could get the delivery address if it is defined.

The DeliveryModuleInterface is deprecated, but is still supported.


InformatiqueProg:
:+1:

roadster31:
This PR is obsolete since the introduction of the MODULE_DELIVERY_GET_POSTAGE event, which passes the address to the module.

PR #2259, opened 2016-09-08
[Coupon] fix cart contains products & cart contains categories conditions

Alban-io:
fix bugs introduced in #2219


bibich:
Thanks @Alban-io (and @TomPradat)

PR #2261, opened 2016-09-08
Cast position_delegate virtual column to number in product loop

### LINKED ISSUE #2260: Product loop "manual" order is sorting positions as strings, not numbers
Thus we get 1, 10, 2, 3, 4, 5, ... instead of 1, 2, 3, ..., 10 when using order="manual"

roadster31:
Fixes #2260 
The manual order is now working properly in the product loop.


bibich:
Thanks

PR #2262, opened 2016-09-13
Use the success_url when deleting a cart_item

Yochima:
This get parameter is sent in default frontOffice template and it is never used

### DIFF
--- a/local/modules/Front/Controller/CartController.php
+++ b/local/modules/Front/Controller/CartController.php
@@ -128,6 +128,12 @@ class CartController extends BaseFrontController
         }
 
         $this->changeViewForAjax();
+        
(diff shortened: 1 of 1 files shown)


local/modules/Front/Controller/CartController.php
```diff
@@ -128,6 +128,12 @@ public function deleteItem()
         }
 
         $this->changeViewForAjax();
+        
+        if (null != $successUrl = $this->getRequest()->query->get('success_url')) {
+            URL::getInstance()->absoluteUrl($successUrl);
```
bibich: `$successUrl = URL::getInstance()->absoluteUrl($successUrl);` instead ?

roadster31:
Thanks !

This PR was manually merged to master, including bibich's fix.

PR #2269, opened 2016-09-17
Fix update sql 2.4.0-alpha1

gillesbourgeat:


bibich:
Never twice without thrice ;)

PR #2272, opened 2016-09-22
Form fields errors are now persisted in ParserContext

roadster31:
Errored forms data are stored in the ParserContext, so that form error information will survive a redirection. But the form fields error informations are not saved in the ParserContext, only the overall form error is persisted.

This is working if no redirection is performed, but causes problems when a redirection is done : when retrieving the form in ParserContext::getForm(), FormValidator::validateForm(), which is used to recreate the fields errors, can't work properly after a redirection, as it uses request data. Thus the form fields error informations are lost.

This PR saves and restores form fields error information along the global form error.

### DIFF
--- a/core/lib/Thelia/Core/Template/ParserContext.php
+++ b/core/lib/Thelia/Core/Template/ParserContext.php
@@ -12,6 +12,8 @@
 
 namespace Thelia\Core\Template;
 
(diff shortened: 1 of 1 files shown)


roadster31:
Travis trigger

PR #2273, opened 2016-09-23
[Coupon] remove html5 number in coupon percentage fragment

Alban-io:


gillesbourgeat:
Hello, why not use the step property 
```html
<input id="coupon-percentage" type="number" class="form-control" name="{$percentage_field_name}" value="{$percentage_value}" step="0.01" placeholder="5">
```

bibich:
A better solution is to use this :

```html
<input id="coupon-percentage" type="number" step="any" min="0" class="form-control" name="{$percentage_field_name}" value="{$percentage_value}" placeholder="5"> 
```

roadster31:
This PR was manually merged to master, with @bibich solution.

Issue #2274, opened 2016-09-28
Bad select/option code source on render_form_field


InformatiqueProg:
{render_form_field form=$form field="post_source"}
a) Tag end ">" of select, textarea, etc. is on a new line
b) option tag are not ended, this is not good
(for example, Javascript if you want to get the text of the option)
![option-not-ended](https://cloud.githubusercontent.com/assets/7933327/18904301/b0dd725c-855d-11e6-8ab2-313923184b8b.PNG)

PR #2275, opened 2016-09-28
bad code source on render_form_field

InformatiqueProg:
Close option tag
remove useless spaces and line return on attributes
See issue #2274


roadster31:
Thanks ! 

This PR was manually merged to master.

PR #2281, opened 2016-10-14
Fixed PECL extentions install script in docker setup

roadster31:
Replaced docker-php-pecl-install  script by the helderco one (https://github.com/helderco/docker-php/blob/master/template/bin/docker-php-pecl-install), to fix this problem : http://thelia.net/forum/viewtopic.php?pid=71132#p71132


jwmarshall:
Confirmed bug and fix. 👍

PR #2283, opened 2016-10-17
Typo which prevents to save the custom-css.less file.

roadster31:
Tjis PR fixes a typo which prevents to save the custom-css.less file.


gillesbourgeat:
:+1:

PR #2287, opened 2016-10-21
Fix for decimals=0 in {format_number} or {format_money}

roadster31:
In  {format_number} or {format_money}, using decimals=0 has no effect, because in core/lib/Thelia/Tools/NumberFormat.php ($decimals == null) is true.

The PR fixes the comparison operator to allow displaying decimal numbers with 0 decimals.


gillesbourgeat:
:+1:

PR #2288, opened 2016-10-24
New CART_ITEM_ID output variable for OrderProduct loop

roadster31:
This variable was missing.


roadster31:
The CartItem ID is only used during Order Creation. In this context, returning the CART_ITEM_ID in the OrderProduct loop is useless.

PR #2291, opened 2016-11-04
Fix currency update

lopes-vincent:
Fix currency change to default where no currency in request.
Change to default only if requested currency not exists.


roadster31:
Could you please rebase ?

lopes-vincent:
Done

roadster31:
Thanks 👍

PR #2296, opened 2016-11-14
Fix for <option> offsets not working in IE and Safari

roadster31:
This PR is a fix for select menus in IE and Safari, where the padding-left styling is not supported, thus resulting in flat selection lists :
img-2016-11-14_11-24-37

A specific Smarty plugin is created, to generate the proper offset using non-breaking spaces.

img-2016-11-14_11-25-51

### DIFF
--- a/templates/backOffice/default/ajax/product-related-tab.html
+++ b/templates/backOffice/default/ajax/product-related-tab.html
@@ -32,7 +32,7 @@
 		                    <select name="folder_id" id="folder_id" class="form-control">
 		                        <option value="">{intl l='Select a folder...'}</option>
 		                        {loop name="folders" type="folder-tree" folder="0" backend_context="1" visible="*" lang="$edit_language_id" return_url=false}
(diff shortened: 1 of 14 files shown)


templates/backOffice/default/ajax/product-related-tab.html
```diff
@@ -32,7 +32,7 @@
 		                    <select name="folder_id" id="folder_id" class="form-control">
 		                        <option value="">{intl l='Select a folder...'}</option>
 		                        {loop name="folders" type="folder-tree" folder="0" backend_context="1" visible="*" lang="$edit_language_id" return_url=false}
-		                            <option value="{$ID}" style="padding-left: {3 + $LEVEL * 20}px">{$TITLE}</option>
+		                            <option value="{$ID}">{option_offset l=$LEVEL label=$TITLE}</option>
```
bibich: I think that the $TITLE will not be escaped

roadster31: True, nice catch :) I've escaped all $TITLE

PR #2308, opened 2016-11-25
Use service instead of instantiate new object

bilhackmac:
Use service to get URL object. It prevent any later call to this service to override request context.


roadster31:
Thanks 👍

PR #2310, opened 2016-11-28
Fix extension of array based loops

bilhackmac:
Arrays are not passed by references. So we need that listeners can set modified array results to event.


roadster31:
Thanks

PR #2325, opened 2016-12-13
Fix lose cart.discount on login

lopes-vincent:
When you create your cart then apply a coupon then login the cart is duplicated but the cart.discount is not duplicated. But the coupon are still in session 
So this PR update the coupon after login to recalculate the cart discount.


bibich:
Thanks

PR #2328, opened 2016-12-21
version.php is now available in https

Soldras:


roadster31:
Thanks 👍

roadster31:
The version check in all Thelia 1 back-offices is now broken. Please re-allow the old http URL 

img-2016-12-23_11-46-22

PR #2330, opened 2017-01-10
Typo

InformatiqueProg:


bibich:
Sorry but you can not change a translation key as some translations could have been done with this key.

PR #2334, opened 2017-01-13
Fix "Class not found" in AbstractArchiver

zorn-v:


roadster31:
Could you please ping to restart the Scrutinizer check ?

roadster31:
Thanks !

PR #2345, opened 2017-01-22
Improved sub-folder and sub-categories count

roadster31:
This PR fixes and improves counting sub categories and sub folders. The number of request is now reduced. 
The visibility status of products or contents is now considered. A new parameter of `countAllProducts()` and `countAllContents()` allow selection of counted products (all / hidden / visible - the default).
The countAllXxxx() methods behavior now complies with the documentation which states that only visible folders/products are counted.

The PR also includes a fix when a category is deleted: category and subcategory products are now deleted by dispatching a `TheliaEvents::PRODUCT_DELETE` instead of calling Product::delete(). This way, database consistency is preserved, specially regarding attributes values.

### DIFF
--- a/core/lib/Thelia/Model/Category.php
+++ b/core/lib/Thelia/Model/Category.php
@@ -3,7 +3,9 @@
 namespace Thelia\Model;
 
 use Propel\Runtime\ActiveQuery\Criteria;
+use Propel\Runtime\Collection\ObjectCollection;
 use Thelia\Core\Event\Category\CategoryEvent;
+use Thelia\Core\Event\Product\ProductDeleteEvent;
 use Thelia\Files\FileModelParentInterface;
 use Thelia\Model\Base\Category as BaseCategory;
 use Thelia\Core\Event\TheliaEvents;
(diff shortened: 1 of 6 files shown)


core/lib/Thelia/Model/Category.php
```diff
@@ -42,22 +44,22 @@ public function getRewrittenUrlViewName()
...
      * /!\ the number of queries is exponential, use it with caution
      *
+     * @param bool|string $productVisibility: true (default) to count only visible products, false to count only hidden
+     *                    products, or * to count all products.
      * @return int
      */
-    public function countAllProducts()
+    public function countAllProducts($productVisibility = true)
     {
         $children = CategoryQuery::findAllChild($this->getId());
```
bibich: It would be better to use `Category::getCategoryTreeIds` function which is lighter as it only deals with categories id and not full objects ?

core/lib/Thelia/Model/Folder.php
```diff
@@ -34,25 +38,25 @@ public function countChild()
...
      * count all products for current category and sub categories
      *
+     * @param bool|string $contentVisibility: true (default) to count only visible products, false to count only hidden
+     *                    products, or * to count all products.
      * @return int
      */
-    public function countAllContents()
+    public function countAllContents($contentVisibility = true)
     {
         $children = FolderQuery::findAllChild($this->getId());
```
bibich: same here with `FolderQuery::getFolderTreeIds`

roadster31: Same here: FolderQuery provides only a filterByFolder()  method.

core/lib/Thelia/Model/Category.php
roadster31: I agree. Unfortunately, ProductQuery provides only a filterByCategory() method, not a filterByCategoryId().

PR #2351, opened 2017-02-01
Changed ModelCriteriaTools to be less restrictive in checking column …

bibich:
…translation instead of row translation.

When using `ModelCriteriaTools` (e.g. in i18n loop) to build I18n propel query and automatically fallback (depends of configuration) on default locale, the old code only checked if there was a translation available (a row present) and not the content of a column that could be NULL. 

It will partially fix the issue #2333.


roadster31:
Thanks !

This PS was manually merged to master.

Issue #2340, opened 2017-01-18
'search_in' is not working in Product loop


roadster31:
Searching is not possible in the product loop. Something like : 

`{loop type="product" name="product_list" search_in="ref,title" search_mode="any_word" search_term={$smarty.get.q}}`

Gives the following exception : 

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''Array') OR CASE WHEN NOT ISNULL(`requested_locale_i18n`.ID)
' at line 4
```

anoziere:
Fixed on both maintained lines: #3472 (merged into `2.6`) and #3473 (merged into `main`, Thelia 3). With `search_mode=any_word`, the word array was bound as a single raw SQL parameter (stringified to `Array`); search clauses are now built with one placeholder per word. The fix lives in the shared i18n search helpers, so the other searchable loops (category, content, folder, brand, sale) benefit as well. Closing manually: `Fixes` keywords only auto-close from the default branch (`main`).

PR #2354, opened 2017-02-09
Fix 'search_in' in different loop see #2340

zzuutt:
Fix 'search_in' in different loop if search_mode is 'any_word'
If we use criteria IN the request, the word is compared to the whole field.
I use REGEX to respect the any_word mode but it also returns the words containing the terms
Otherwise it would be necessary to write :
$searchTerm = '[[.space.]]('.str_replace(' ','|', rtrim($searchTerm)).')[[.space.]]';


roadster31:
6 new issues in scrutinizer, could you please fix this ?

zzuutt:
désolé mais je ne comprends pas l'erreur
que dois je faire ?

Issue #2358, opened 2017-02-22
User info not updated in front if user is connected


TomPradat:
If an admin changes user information when the user is connected, the information are not updated in front because the customer is fetch from the security context which still have the previous information.

roadster31:
This information is stored in the user session, which is not updated at each request.

Maybe we should perform this update at each request ?

TomPradat:
No it's too much resourceful for our need, let's just change how the customer controller gets the user info for the route account/update, i'm making a PR

PR #2359, opened 2017-02-22
Create a query to get customer information instead of using session

TomPradat:
Resolve issue #2358


roadster31:
I suggest to implement this check in the getCustomerUser() method of SecurityContext, and cache the result for the current request, and do the same for the admin user, as discussed in #2089.

What do you think ?

roadster31:
@TomPradat any update ?

TomPradat:
As said in #2089, why not deleting the session of the user when changed or deleted ? We have or should have events fired when a user changed. Let's just kill the session if the user is deleted or the user is an admin. Also with this pr we make sure that is user change information he gets the right information in his profile view

zorn-v:
For "delete session of the user" you must find file in local/sessions with that session on server and delete it. 
And just imagine a module that keep sessions in another place (in some db for example).

You must store all sessions (ids) of all users somewere to properly destroy them.
Do you really think it is good sollution ?

TomPradat:
Well we could have a table with the session ids associated to users and have an interface to destroy sessions so that you can store the sessions wherever you want.

zorn-v:
> Do you really think it is good sollution ?
AFAIR you also must invalidate it on client for "things go right"

TomPradat:
I don't understand your last comment. But i think this could be a good solution instead of having more and more checks on every request that are usefull 1 out of 10 000 or so. This is the most logical way to do it from my point of view. This is the delete/change user action which should lead to disconnecting the user, not checking at each user request if he is still in the database and hasn't changed.

Issue #2347, opened 2017-01-26
Removing an attribute from a product template do not delete related combinations


roadster31:
When an attribute is deleted or removed from a product template, the related combinations are not deleted from the products which uses this template.

mbruchetpro:
Hi ! 
With the latest version of thelia, do you still have the same issue ? 
Because when I test it on this version, it remove all attribute link with the product I test.

PR #2363, opened 2017-03-01
Improved products templates features

roadster31:
This PR improves the product templates features:
1) When a product's template is modified, only the obsolete attribute combinations are deleted, this is much better than deleting all attribute combinations.
2) When an attribute is removed from a template, all related products are also updated, and obsolete attribute combinations are removed.
3) When a feature is removed from a template, the related products are updated, and the feature is removed from the product.
4) The template management page now displays the related products and categories, making finding which products or categories uses the template easier.
5) It is now possible to duplicate a product template
6) Product and Category loops have now a `template_id` parameter, to filter results on one ore more template IDs


roadster31:
Fixes #2347

PR #2364, opened 2017-03-08
HTTPS on iframe src

InformatiqueProg:
If we leave the iframe src on HTTP, it's wok only if site is called on HTTP
With the iframe src is on HTTPS, it's work for both


roadster31:
Thanks.

PR #2365, opened 2017-03-15
grammar edit: this hooks -> these hooks

acobster:


acobster:
Last two commits address https://github.com/thelia/thelia/issues/2366

templates/backOffice/default/hooks.html
```diff
@@ -510,7 +510,7 @@
 <script id="discover-missing" type="text/html">
     <table class="table table-striped table-condensed">
         <caption class="clearfix">
-            {intl l='Your template does not support this hooks'}
+            {intl l='Your template does not support these hooks'}
```
roadster31: Instead of changing the sentence in the html page, which implies changing all translation files, it is better to change only the English translation in the en_US.php translation file.

templates/backOffice/default/hooks.html
```diff
@@ -531,7 +531,7 @@
         </tr>
         <% }); %>
         <tr>
-            <td colspan="5"><i class='glyphicon glyphicon-arrow-up'></i> {intl l="Deactivate this hooks"}</td>
+            <td colspan="5"><i class='glyphicon glyphicon-arrow-up'></i> {intl l="Deactivate these hooks"}</td>
```
roadster31: Please see comment at line 513.

templates/backOffice/default/hooks.html
roadster31: Instead of changing the sentence in the html page, which implies changing all translation files, it is better to change only the English translation in the en_US.php translation file.

templates/backOffice/default/hooks.html
roadster31: Please see comment at line 513.

acobster:
OK, updated (rebased)

roadster31:
Thank you !

PR #2368, opened 2017-03-23
Fix Cannot use Thelia\Model\Module as Module because the name is already in use

remadeinfrance:
Remove "use Thelia\Model\Module" en Delivery loop

exemple : 
{loop type="delivery" name="shipping_module" backend_context="on"}
{$TITLE}
{/loop}


roadster31:
I can't reproduce the problem. Could you please give more details ?

roadster31:
Any update  ?

PR #2371, opened 2017-04-02
Updated abandoned deps

zorn-v:


zorn-v:
Seems need to update (generate) composer.lock with php 5.4 
Currently generated with php 7.0

PR #2376, opened 2017-04-11
Fix manual sorting in product loop

zorn-v:
First of all there is no sense in that case, coz `product_category.position` column can not be NULL.
Second - with this construction there is an error (at least with php x64). 
Propel set qoutes around that value: `THEN '9223372036854775807' ELSE` and mysql set type of `position_delegate` as string and sort order becomes wrong. Like this `1 10 11 ... 18 19 2 20`
You can check it in backoffice by sorting products in category by position with products count over 20 in that cat.

### DIFF
--- a/core/lib/Thelia/Core/Template/Loop/Product.php
+++ b/core/lib/Thelia/Core/Template/Loop/Product.php
@@ -609,10 +609,7 @@ class Product extends BaseI18nLoop implements PropelSearchLoopInterface, SearchL
             ;
         }
     
-        $search->withColumn(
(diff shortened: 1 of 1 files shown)


zorn-v:
up

roadster31:
I can't reproduce the problem on the current master version.

> product_category.position column can not be NULL
We have to take care of this specific case, one never knows.

zorn-v:
https://github.com/thelia/thelia/pull/2261 - seems fix it too. 
I've got that on 2.3.3 on which I'm develop now. 
I do not pay attention that it differ in master, sorry.
> We have to take care of this specific case

On which case are you talking ? That column cannot be null on DB level. 
You just cannot insert or update that column to null
Anyway good to know that similar patch already in master

PS: Can you please merge https://github.com/thelia/thelia/pull/2378 
Just a little late to push it to original PR

PR #2379, opened 2017-04-20
Set default tax rule if it not defined in create product event

zorn-v:


core/lib/Thelia/Action/Product.php
```diff
@@ -80,6 +80,12 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
      */
     public function create(ProductCreateEvent $event)
     {
+        $defaultTaxRuleId = null;
+        $defaultTaxRule = TaxRuleQuery::create()->findOneByIsDefault(true);
+        if ($defaultTaxRule) {
```
roadster31: Better code style is :
```
if (null !== $defaultTaxRule = TaxRuleQuery::create()->findOneByIsDefault(true))  {
...
}
```

zorn-v: No problem

roadster31:
Thanks.

PR #2383, opened 2017-05-04
Call built-in setlocale before session start

zorn-v:
Useful for smarty `date_format` with format like `%A` for example


zorn-v:
Oops, forgot about thelia smarty functions

Issue #2069, opened 2016-03-25
Deleting a product with a free text feature value using SQL request or Product::delete() is breaking the feature behaviour.


roadster31:
When a product is not deleted via `Action\Product\delete()` (eg. with a simple SQL request, or with a call to `Product::delete()`), if it has a free text feature, the related Feature Availability is not deleted, and it causes the Feature to be changed from a free text feature to a list feature. (see  #2061, which is not completely fixed by #2063)

This is the case when a category is deleted, as the product is automatically deleted by the database (delete cascade...)

To solve this problem, we have to change the database structure to manage the free_text_value flag in the `feature_av` table instead of the `feature_product` table.
However, this will not completely solve the problem, as obsolete lines will be left in the `feature_av` table.

roadster31:
Fixed by #2392 (hopefully...)

PR #2392, opened 2017-05-27
Feature free text value refactoring

zorn-v:
That field (free_text_value) makes frustrating some people (like me) just by its name and type. But later you think about i18 etc. 
Any "feature" must be consistently with other system parts I think.
Goal of thist PR is remove that damn field and replace it with `is_free_text`. 
Anyway it'll be just deprecated, until 2.5.0 i think...


zorn-v:
There was copy&paste bug https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Core/Template/Loop/FeatureValue.php#L103

zorn-v:
Feature tax amount git not properly https://github.com/thelia/thelia/blob/230696257aaa5e684449f6f652ea5ccbe337b074/core/lib/Thelia/TaxEngine/TaxType/FeatureFixAmountTaxType.php#L49

zorn-v:
Just notice https://github.com/thelia/thelia/blob/230696257aaa5e684449f6f652ea5ccbe337b074/core/lib/Thelia/Core/Template/Loop/Product.php#L471

zorn-v:
We can not reach request in tax engine. And so we can't access to current locale. Let's use default one.

zorn-v:
Seems to be done

roadster31:
Thanks for this nice job ! Your help is much appreciated 👍 

Just two points :

1) Could you please override in core/lib/Thelia/Model/FeatureProductQuery.php and core/lib/Thelia/Model/FeatureProduct.php all method related to the free_text_value column, and deprecate them, so thant plugins writers realize the deprecation of the column ?

2) Could you please create a specific PR to change the  version to 2.4.0-alpha2 ? I can also do it myself, if you prefer.

zorn-v:
1. I did not override FeatureProduct::setFreeTextValue because it used in create/update action. But getFreeTextValue also used in loop... Well, will be added.
2. Version change needs to add `is_free_text` to DB (for update script work). And without that field other changes does not sense. I dont know how decouple version change from other changes. If you have any thoughts - do it.

roadster31:
Overriding all FreeTextValue related methods allow us to deprecate them, without changing method behaviour. For example for getFreeTextValue() : 

```
    /**
     * @inheritdoc
     * @deprecated since version 2.4.0, to be removed in 3.0.
     *                      Please use  getIsFreeText() instead
     */
    public function getFreeTextValue()
    {
        return parent::getFreeTextValue();
    }
```

core/lib/Thelia/Model/FeatureProductQuery.php
```diff
@@ -16,5 +17,14 @@
...
+    /**
+     * @deprecated
+     */
+    public function filterByFreeTextValue($freeTextValue = null, $comparison = null)
+    {
+        $bt = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        Tlog::getInstance()->warning(sprintf('Using `free_text_value` is deprecated. Use `is_free_text` instead. Used in %s:%d', $bt[0]['file'], $bt[0]['line']));
+        return parent::filterByFreeTextValue($freeTextValue, $comparison);
+    }
 }
```
roadster31: Please deprecate also all *FreeTextValue() methods, such as orderByFreeTextValue(), findOneByFreeTextVaue(), and so on.

zorn-v:
findBy and findOneBy calls filterBy method, which already deprecated.

core/lib/Thelia/Model/FeatureProductQuery.php
zorn-v: findBy and findOneBy calls filterBy method, which already deprecated.

roadster31: True. But I insist on providing also PHPDocs for these methods.

zorn-v: Ok

core/lib/Thelia/Action/Product.php
```diff
@@ -667,6 +667,7 @@ public function updateFeatureProductValue(FeatureProductUpdateEvent $event)
             // If it's a free text value, create a FeatureAv to handle i18n
             if ($event->getIsTextValue() === true) {
                 $featureProduct->setFreeTextValue(true);
```
roadster31: To prevent the use of deprecated methods in Thelia code, I suggest that `setIsFreeText() ` call `setFreeTextValue()`

core/lib/Thelia/TaxEngine/TaxType/FeatureFixAmountTaxType.php
```diff
@@ -46,7 +47,9 @@ public function fixAmountRetriever(Product $product)
             ->findOne();
 
         if (null !== $query) {
-            $taxAmount = $query->getFreeTextValue();
+            if ( (1 == $taxAmount = $query->getFreeTextValue()) && !is_null($query->getFeatureAvId()) ) {
```
roadster31: Please use `getIsFreeText()` instead of  the depracted method `getFreeTextValue()`

core/lib/Thelia/Model/FeatureProductQuery.php
```diff
@@ -16,5 +18,60 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     * @deprecated since version 2.4.0, to be removed in 3.0.
+     *                      Please use  findByIsFreeText() instead
+     */
+    public function findByFreeTextValue($free_text_value)
+    {
+        return parent::findByFreeTextValue($free_text_value);
```
roadster31: Could you please also log a warning, as in other deprecated methods ?

core/lib/Thelia/Model/FeatureProductQuery.php
```diff
@@ -16,5 +18,60 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     * @deprecated since version 2.4.0, to be removed in 3.0.
+     *                      Please use  findOneByIsFreeText() instead
+     */
+    public function findOneByFreeTextValue($free_text_value)
+    {
+        return parent::findOneByFreeTextValue($free_text_value);
```
roadster31: Could you please also log a warning, as in other deprecated methods ?

roadster31:
Thanks for the update.

Could you please search occurrences of *FreeTextValue in the Thelia code, and replace them with the new *IsTextValue methods ? There is still a bunch of them in the code (e.g. faker.php, Model\Product.php, Action/ProductTest.php and others)

roadster31:
I just remembered this comment https://github.com/thelia/thelia/issues/2069, about the issue #2061 .

What about moving the IsTextValue flag in the feature_av table, to provide a better (but not yet definitive) solution to #2061 ?

core/lib/Thelia/Action/Product.php
zorn-v: Good idea. And also call `setIsFreeText` in `setFreeTextValue` for BC

core/lib/Thelia/TaxEngine/TaxType/FeatureFixAmountTaxType.php
zorn-v: Its for BC. Just imagine manually filled DB (or via some import) with using of this tax type.
Seems to need add php upgrade script for converting old `free_text_values` to "default locale i18n"  `feature_avs`

zorn-v:
Don't get how moving that field from one table to another helps with that issue...

core/lib/Thelia/Model/FeatureProductQuery.php
zorn-v: Deprecation already logged in filterByFreeTextValue. But it show wrong place of caller (propel internals)... Ok, I'll add additional logs for that methods.

core/lib/Thelia/TaxEngine/TaxType/FeatureFixAmountTaxType.php
zorn-v: Its seems that already was in 2.2.0-beta1.php
Let's count that no one have `NULL` in `feature_av_id` then ?

zorn-v:
Delete of category also deletes all products related to it. And product delete all free text features. There is also one case when delete some of them from DB manually. 
In that point trigger will be great as it not Thelia fault.

PS. Such category deletion cause big slowdown on delete category with big children tree and many products in it.

tests/phpunit/Thelia/Tests/Coupon/Type/FreeProductTest.php
```diff
@@ -19,6 +19,7 @@
 use Thelia\Condition\Implementation\MatchForTotalAmount;
 use Thelia\Condition\Operators;
 use Thelia\Coupon\FacadeInterface;
+use Thelia\Coupon\Type\FreeProduct;
```
zorn-v: Fixed php 5.5 test failed in some cases

roadster31:
Using a DB trigger would be safer, this is a good idea. 

Btw, could you please rebase your branch ?

zorn-v:
What do you think about defining `feature_av_id` as `NOT NULL` and rewriting loop and feature tax engine according it ?

zorn-v:
There is some bug needs to inspect. 
To reproduce it
1. Create "feature" tax with "IT" language
2. Add that tax to some tax rule
3. Set some values for that "tax feature" in EN and IT locales on product with tax rule from 2
4. Remove value from IT locale
And you got 
```
Unable to execute UPDATE statement [UPDATE `feature_product` SET `FEATURE_AV_ID`=:p1, `UPDATED_AT`=:p2 WHERE feature_product.ID=:p3]
```
```
SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`thelia_dev`.`feature_product`, CONSTRAINT `fk_feature_prod_feature_av_id` FOREIGN KEY (`feature_av_id`) REFERENCES `feature_av` (`id`) ON DELETE CASCADE)
```

zorn-v:
Item 1 and 2 not needed to reproduce a bug.
Just fill free text in two locales and try to remove one of them.

I think I understand what's going on and try to fix it

core/lib/Thelia/Install/Database.php
```diff
@@ -129,6 +129,15 @@ protected function prepareSql($sql)
...
         $sql = trim($sql);
+        preg_match_all('#DELIMITER (.+?)\n(.+?)DELIMITER ;#s', $sql, $m);
+        foreach ($m[0] as $k => $v) {
+            if ($m[1][$k] == '|') {
+                throw new \RuntimeException('You can not use "|" as delimiter: '.$v);
+            }
+            $stored = str_replace(';', '|', $m[2][$k]);
+            $stored = str_replace($m[1][$k], ";\n", $stored);
+            $sql = str_replace($v, $stored, $sql);
+        }
```
roadster31: Nice catch :)

zorn-v: I'm always try to adapt to existing code
If existing code worth it of cause :)

zorn-v: DELIMITER things not actually needed (DELIMITER is mysql client "reserved word") but I thought that sql should import normal thru `mysql < thelia.sql`
But there was no logic for import such things thru PDO.

zorn-v: Maybe assign ugly `$m[2][$k]` to meaningfull variables names for clarity ?

roadster31: One letter variables names are not a good practice.

zorn-v: Yep, but `$matches[1][$k]` does not say something too

zorn-v:
Up. Is it was useless work ?

roadster31:
No. 

I've not completely reviewed this big PR. Plesase be patient.

zorn-v:
Sorry

roadster31:
Thanks for this nice work :)

PR #2397, opened 2017-05-31
Fixed tests

zorn-v:


zorn-v:
Can anyone explain what 'Must be revisited. Tax Rules do not work like this.' mean ? 
It's just API controller tests. Is it suggest rewriting controller or what ?

zorn-v:
Done

zorn-v:
I just catch. Tests actually just for for show ?

PS. BTW some tests  undetermined and depends on "clean import". but sometimes mysql `RAND` plays a cruel joke and test not catch address2 for example (in smarty module test). And so on.

roadster31:
I'm not sure to understand what you mean. 

Tu peux écrire en français si tu veux.

zorn-v:
>Tu peux écrire en français si tu veux.

Sorry I'm Russian :smiley: 
I meant that not such attention invested in tests. Thats all.

zorn-v:
That tests only test API controllers. Is that controllers is not changed in 2.3 ?
If they WAS changed, that tests actually say that controllers changed properly. 
Or you mean that there is need to add some data to test with states ?

One more time - that tests DOES NOT actually check some tax behaviour, they test only response codes from API controller on create/update/delete/createWithInvalidData actions.

roadster31:
> Or you mean that there is need to add some data to test with states ?
Tax rules are now attached to a countries/states configuration. This is not implemented in the tests.

tests/phpunit/Thelia/Tests/Api/TaxRuleControllerTest.php
```diff
@@ -230,12 +226,10 @@ public function testUpdateTaxRule($taxRuleId)
 
     /**
      * @param $taxRuleId
-     * @depends testCreateTaxRule
+     * @depends testUpdateTaxRule
```
roadster31: Why did you changed this ?

zorn-v: To explicit specify create->update->delete test flow.
Actually that does not change anything coz delete test follow after update test, but explicit better than implicit :smile: 
And I think that was just copy/paste issue (`testCreateTaxRule` instead of `testUpdateTaxRule`)

zorn-v:
> Tax rules are now attached to a countries/states configuration. This is not implemented in the tests.

By saying that you assert that it is not implemented in API part, not in that tests. And if it true, test not good place to put notice for it. It must be in API controller response then (if API controller not work properly of cause).

Take a look carefully what that tests doing.
And finally check that controller https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Controller/Api/TaxRuleController.php#L143
There is not reference to states at all...

zorn-v:
Maybe you mean to change that test for not using "compat" return value there https://github.com/thelia/thelia/blob/master/core/lib/Thelia/Action/TaxRule.php#L156 ?
I really don't understand what mean "Tax rules are now attached to a countries/states configuration" and how that tests must implement it, considering that it just make request to `/api/tax-rules` url

PR #2411, opened 2017-08-17
Remove required on pses weight field

lopes-vincent:
I think there is no reason to have a required here because pse.weight can be NULL.


roadster31:
Could you please rebase ?

lopes-vincent:
Done

roadster31:
Manually merged.

PR #2412, opened 2017-08-18
Fix syntax error in CarouselLoop

Paphos:


roadster31:
Nice catch, thank you

PR #2417, opened 2017-08-23
Add area delivery module check

marsender:
In getCheapestDelivery, the country area must be checked or we may select an incorrect delivery.
Most common case is probably the localdelivery, with a price of "0" for example. This delivery price is selected even if the customer country is not valid.
So I have added the same check as in the delivery selection file:
/vendor/thelia/core/lib/Thelia/Core/Template/Loop/Delivery.php
Thanks


core/lib/Thelia/TaxEngine/TaxEngine.php
```diff
@@ -128,8 +128,10 @@ public function getDeliveryCountry()
                     $this->taxState = $currentDeliveryAddress->getState();
                 } else {
                     $customerDefaultAddress = $customer->getDefaultAddress();
-                    $this->taxCountry = $customerDefaultAddress->getCountry();
-                    $this->taxState = $customerDefaultAddress->getState();
+                    if (isset($customerDefaultAddress)) {
+											$this->taxCountry = $customerDefaultAddress->getCountry();
```
roadster31: Could you please use spaces instead of tabs for indentation ?

marsender: Yes of course. I've replaced tabs with spaces

roadster31:
There are some problems with the new Travis CI configuration which have been fixed by #2422. Could you please rebase ?

marsender:
The rebase is done. It's ok.

roadster31:
Thanks 👍

PR #2423, opened 2017-09-03
Removed junk file

roadster31:
The commits-2.3.3.ods file is not a project file, neither test.html nor the HookCookie directory.


zorn-v:
There is also `Submodule HookCookies added at 85c49e ` with that commit

zorn-v:
There was strange commit. Look at `templates/frontOffice/default/test.html`
https://github.com/thelia/thelia/commit/1d5c2094f36174c47644e3b65033418cf07f9de7

roadster31:
True, I have to remove this file also.

PR #2436, opened 2017-10-07
Add remove function on loop result collection

InformatiqueProg:
TheliaEvents::LOOP_EXTENDS_PARSE_RESULTS

We can on loop results : 
- Add row
- Change each row values

but we can't remove row that's why i propose this PR


roadster31:
Thanks !

InformatiqueProg:
you're welcome!

PR #2437, opened 2017-10-12
correct order name differences between definition and buildModelCriteria

InformatiqueProg:


roadster31:
Thanks.

Could you please ensure backward compatibility, by leaving the old `alpha` and `alpha_reverse` values ?

InformatiqueProg:
i have checked before do this PR, all loop "area" don't use order argument

- I'll add `alpha` for backward compatibility
- but why add `alpha_reverse` it wasn't in previous `EnumListType` ?

roadster31:
> but why add alpha_reverse it wasn't in previous EnumListType ?
You're right, sorry.

InformatiqueProg:
Done!

roadster31:
Thanks !

InformatiqueProg:
You're welcome! 😉

PR #2441, opened 2017-10-13
Fix product cloning : clone additionnal categories

Paphos:
The additional categories were missing in the product cloning. Fixed.


roadster31:
Thanks 👍

Issue #2446, opened 2017-10-26
Version 2.3.4 not installable


zorn-v:
First - there is no thelia-project for 2.3.4 
If you update to it you MUST create favicon.png and logo.png
And you can not do it thru admin panel coz it will throw exception too.

This all smells like a shit.

roadster31:
> This all smells like a shit.

Such comments are not welcome here.

zorn-v:
Sorry for straightness.
But I feel ssomething like "there was perfect framework for shop, but now %$!@#%$"

zorn-v:
I know that you are the one. But on thats things (like you cannot just run it without some doings) I'm disappointed

zorn-v:
BTW. thelia/thelia-project is out of date.
Can you give me rights for "fix" it ?

roadster31:
The 2.3.4 release of thelia-project has just been created.

Jingle bells.

zorn-v:
I know, and because of it I'm asking.
Got 2.3.4 you shoud install, change version in composer, MANUALLY (coz admin panel does not work for this) upload favicon.png and logo.png (paths you should get from exception messages). 
Wow You installed 2.3.4

roadster31:
Explaining your issue like this is fine. At least better than just ranting and talking of 'shit'.

Whatever. Could you please open a new issue related to this specific problem ?

PR #2416, opened 2017-08-22
Allow changing favicon and logo in back-office

Paphos:
- Favicon and logo are no longer related to the template and can be changed in the back-office in Configuration / Store.
- The file name of the logo and favicon is stored in database in the "config" table with the keys "favicon_file" and "logo_file".
- Default Thelia favicon and logo are stored in local/media/images/store and used if no name is provided in the "config" table.
- Smarty plugins {logo} and {favicon} are used to get the image path.

PR #2448, opened 2017-10-27
Fix bug when store images are missing

Paphos:
(Fix issue #2446)
My previous PR (#2416) indeed threw an exception on the "Config > Store management" page when the images are not found.
It didn't cause problems when installing Thelia 2.3.4 from scratch as the default three images (favicon.png, logo.png and banner.jpg) were copied during the installation. But this was a problem when updating to Thelia 2.3.4 as these images are not copied.
This PR just ensures that the smarty block {local_media} doesn't throw exception anymore when the image are missing, instead the variable $MEDIA_URL will be empty.


roadster31:
Thanks.

PR #2450, opened 2017-11-09
Add pagination in the back-office translation tool

Paphos:
=> /admin/configuration/translations


roadster31:
Something seems wrong when all strings have been translated, and the "View only missing translations" is checked.

Normal display :
image

With this PR :
image

Could you please check this ?

lopes-vincent:
Pagination will not work with the current translation system we need to refactor it before add pagination

PR #2453, opened 2017-11-18
Added exclude_code argument

roadster31:
This PR adds the `exclude_code` argument to delivery and payment loops, to allow excluding a module using its code instead of its ID


core/lib/Thelia/Core/Template/Loop/BaseSpecificModule.php
```diff
@@ -28,6 +28,7 @@
  * {@inheritdoc}
  * @method int getId()
  * @method int[] getExclude()
+ * @method int[] getExcludeCode()
```
gillesbourgeat: Hello,
The typage is `string[]` not `int[]`

roadster31: Nice catch :) Thanks.

PR #2457, opened 2017-11-29
Prevent useless data copy in export

roadster31:
The export system is perfoming a useless array copy, just for testing purposes. This PR is removing this copy.


roadster31:
This PR does not help to reduce memory usage.

PR #2469, opened 2018-01-30
New feature: template inheritance

roadster31:
This PR introduces a new feature, the template inheritance. You can now create a new template, and redefine only the required files instead of copying the whole template.

To do so, you have to create in your template directory a template descriptor file, named `template.xml`, which contains some information about your template: 
```
<?xml version="1.0" encoding="UTF-8"?>
<template xmlns="http://thelia.net/schema/dic/template"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://thelia.net/schema/dic/template http://thelia.net/schema/dic/template/template-1_0.xsd">
    <descriptive locale="fr">
        <title>Ce template back-offcie hérite de "default"</title>
    </descriptive>
    <descriptive locale="en">
        <title>This back-offcie template inherits from the "default" template</title>
    </descriptive>
    <parent>default</parent>
    <languages>
        <language>en_US</language>
        <language>fr_FR</language>
    </languages>
    <version>1.0.0</version>
    <authors>
        <author>
            <name>Thelia team</name>
            <company>thelia.net</company>
            <email>contact@thelia.net</email>
            <website>thelia.net</website>
        </author>
    </authors>
    <thelia>2.4.0</thelia>
    <stability>prod</stability>
</template>
```
The schema is very similar to the modules descriptor. The important element here is `parent`, which identifies the parent template, which should be an existing template of the same type (fo, bo, pdf, mail) as the child template. In our example, the new template inherits from the `default` template.

All template files are inherited, including the assets, the translations and the module overrides.

However, you can define specific translations for your template, and use them by declaring a default translation domain in your template files, for example `{default_translation_domain domain='bo.your_template_name'}`, or using the `domain `(abbreviated `d`) parameter of the `intl `Smarty function :` {intl l='Edit a customer' d='bo.your_template_name'}`

You can also use your own assets, as in a regular template. Use a `{declare_assets directory='your_asset_directory'}` if your CSS or JS references relative resources, so that they could be copied in the web/assets directory.

Additionally, you can also override the assets of your parent if you're using the same asset directory structure. For example, if you want to override the `assets/css/styles.css`  of your parent, create a  `assets/css/styles.css` in your template, and it will override the parent's one.

 :warning: Don't forget to clear the cache when manipulating templates, as template and/or assets information may be cached by Thelia.

The ParserInterface interface has now 2 new methods :
- `pushTemplateDefinition(TemplateDefinition $templateDefinition, $fallbackToDefaultTemplate)` to set a new template definition, while saving the current one
- `popTemplateDefinition()` to restore a previously pushed template

These methods are useful to temporary set a template (for sending a mail, for example), and then restore the current front or back template.

### DIFF
--- a/core/lib/Thelia/Action/Lang.php
+++ b/core/lib/Thelia/Action/Lang.php
@@ -26,6 +26,7 @@ use Thelia\Core\Event\Lang\LangToggleVisibleEvent;
 use Thelia\Core\Event\Lang\LangUpdateEvent;
 use Thelia\Core\Event\TheliaEvents;
 use Thelia\Core\HttpFoundation\Session\Session;
+use Thelia\Core\Template\Exception\TemplateException;
 use Thelia\Core\Template\TemplateHelperInterface;
 use Thelia\Core\Translation\Translator;
 use Thelia\Form\Lang\LangUrlEvent;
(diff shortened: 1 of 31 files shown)


InformatiqueProg:
👍

PR #2470, opened 2018-01-30
Fix missing tax states from loops

marsender:
This PR allows states to be taken into account in tax rules.
Without this, if you are in a country like US with multiple states, the state filter is not used and the tax is added for each state.
The problem occurs for any tax rules applied to a country with states.
Thanks

### DIFF
--- a/core/lib/Thelia/ImportExport/Export/Type/MailingExport.php
+++ b/core/lib/Thelia/ImportExport/Export/Type/MailingExport.php
@@ -13,6 +13,7 @@
 namespace Thelia\ImportExport\Export\Type;
 
 use Thelia\ImportExport\Export\AbstractExport;
+use Thelia\Model\Map\NewsletterTableMap;
 use Thelia\Model\NewsletterQuery;
 
 /**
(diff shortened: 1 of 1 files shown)


roadster31:
Could you please give a description of your PR ?

marsender:
This PR allows states to be taken into account in tax rules.
Without this, if you are in a country like US with multiple states, the state filter is not used and the tax is added for each state.
The problem occurs for any tax rules applied to a country with states.
Thanks

roadster31:
In this PR, you're changing the signature of several public methods in core/lib/Thelia/Model/ProductSaleElements.php and core/lib/Thelia/Model/Product.php, which causes à BC break.

Maybe you can add additional method parameters at end of parameter list, with a default value, or use new methods (e.g. getTaxedPriceWithState) and keep old ones with the same signature.

What do you think ?

marsender:
Hi,
I saw that and have added this change: https://github.com/thelia/thelia/pull/2470/commits/a7d793c1298f40b7ab006ba1a569d8999bf29a7e
in order to set a null default value: State $state = null
But sorry, I forgot to add it  in the PR

roadster31:
Hi,

Even with your last changes, the signature of the method is modified, as parameters are not in the same order;

`public function getTaxedPrice(Country $country, $price)` becomes `public function getTaxedPrice(Country $country, State $state = null, $price = 0)`.

The following signature don't introduce a compatibility break, as method parameters appear in the same order, and the new $state parameter has a default value : 

`public function getTaxedPrice(Country $country, $price, State $state = null)`.

core/lib/Thelia/Handler/ExportHandler.php
```diff
@@ -216,7 +234,10 @@ public function export(
             }
 
             // Finalize archive
-            $event->getArchiver()->add($filePath)->save();
+            // -DC- Set export file name inside zip
+            //$event->getArchiver()->add($filePath)->save();
```
roadster31: Please do not left the code you replaced as a comment, just remove it.

core/lib/Thelia/Handler/ExportHandler.php
```diff
@@ -216,7 +234,10 @@ public function export(
             }
 
             // Finalize archive
-            $event->getArchiver()->add($filePath)->save();
+            // -DC- Set export file name inside zip
```
roadster31: What `-DC-` means ? (global remark)

marsender: Sorry, these are my initials.
I shouldn't have put them on.

PR #2471, opened 2018-02-02
update imagine, add base64 and blur option for image loop

Lucanis:
- Bump imagine from 0.6 to 0.7 to avoid future problem with recent version of imagick
- Add  blur effect option on image (works best with imagick)
- Add optional param to get the image in base64


roadster31:
Thnaks for this PR !

Could you please update the documentation of the image loop ?

Lucanis:
https://github.com/thelia/thelia.github.io/pull/261

roadster31:
Thanks !

PR #2500, opened 2018-05-03
Fix newsletter unsubscribe/subscribe

yunai39:
The value of the newsletter input is true if there is a newsletter for that email, but it does not care if the newsletter is unsubsribed or not

In the update action of the newsletterAction, unsubscribe is never change. Therefore, a customer cannot subscribe/unsubscribe to the newsletter. Has update is only called when newsletter is true, I setted unscubscribed to false


core/lib/Thelia/Action/Newsletter.php
```diff
@@ -79,6 +79,7 @@ public function update(NewsletterEvent $event)
             $nl->setEmail($event->getEmail())
                 ->setFirstname($event->getFirstname())
                 ->setLastname($event->getLastname())
+                ->setUnsubscribed(0)
```
roadster31: ->setUnsubscribed(false) would be better :)

PR #2502, opened 2018-05-22
Added category and brand ID in sidebar hooks

roadster31:
In brand and category sidebars, the category ID and the brand ID were missing.


gillesbourgeat:
Why do you want to delete file `templates/frontOffice/default/includes/categories-filters.html` ?

roadster31:
It is not user anywhere, and its content is not usable as-is.

PR #2508, opened 2018-07-04
Free the footer's hooks

InformatiqueProg:
If you use Hook is that you want something different...
why constrain the render ?


InformatiqueProg:
I'm agree with you, this introduce a BC.
After all, i not sure we need to make a BC for this change...

roadster31:
Okay, let's close it, then.

PR #2511, opened 2018-07-05
Fix a path generation issue in Propel class generators

### LINKED ISSUE #2490: Model is regenerated in core/lib/Thelia/Thelia/Model instead of core/lib/Thelia/Model
When the cache is cleared, Thelia 2.4.0 alpha2 regenerates the Thelia model in the wrong directory, core/lib/Thelia/Thelia/Model  instead of core/lib/Thelia/Model

bcbrr:
The `endPath` arguments passed to `Symfony\Component\Filesystem\Filesystem::makePathRelative()` had a double path separator.
For the path that has a `..` this caused an issue with path generation: the `..` segment was not properly applied because of this double separator.

Fixes #2490

### DIFF
--- a/core/lib/Thelia/Core/Propel/Generator/Builder/Om/Mixin/ImplementationClassTrait.php
+++ b/core/lib/Thelia/Core/Propel/Generator/Builder/Om/Mixin/ImplementationClassTrait.php
@@ -31,7 +31,7 @@ trait ImplementationClassTrait
 
         return $fs->makePathRelative(
             THELIA_CACHE_DIR
-            . (defined('THELIA_PROPEL_BUILDER_ENVIRONMENT') ? ('/' . THELIA_PROPEL_BUILDER_ENVIRONMENT) : '')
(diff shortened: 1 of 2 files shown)


roadster31:
:+1:

roadster31:
I'm working with windows (OK, I know), and I can see that there's a mix with \ and / in path, meaning that sometimes "/" is used to build path instead of the DS constant.

bcbrr:
I changed these two classes to use `DS`.

However, `makePathRelative` is converting Windows path separators to UNIX path separators, but does not convert back when returning. Should this be done ? (There is four uses of this method in Thelia : these 3, and one in `ModuleGenerateCommand`).

I also used `/` separators in #1975, should I rework it to change everything to `DS` ?

roadster31:
> I also used / separators in #1975, should I rework it to change everything to DS ?
I did not review this PR. But IMO, if you're working on file system paths, you should use DS everywhere.

PR #2539, opened 2018-10-25
A missing loop type will throw an error in dev mode only

roadster31:
When a loop type is not found (e.g. `{loop type="some-unknown-loop-type" ...}`, an error is thrown in development mode. In production mode, and empty result is assumed. For {count} loops, a count of 0 is returned. An error message is added to the Thelia log.

This way, when a module is disabled, the front office is not broken if module's loops are used in the templates.


roadster31:
Included in #2542

PR #2547, opened 2018-10-29
BO new buttons integration

gillesbourgeat:
## small screen
capture du 2018-10-29 21-10-22
## large screen
capture du 2018-10-29 21-10-39


roadster31:
Top :+1:

PR #2574, opened 2018-11-16
A missing hook will throw an error in dev mode only

gillesbourgeat:


local/modules/TheliaSmarty/Template/Plugins/Hook.php
```diff
@@ -287,9 +288,17 @@ public function processForHookBlock($params, $content, $smarty, &$repeat)
...
-                throw new \InvalidArgumentException(
+                $exception = new \InvalidArgumentException(
                     $this->translator->trans("Related hook name '%name' is not defined.", ['%name' => $rel])
                 );
+
+                if ($this->debug) {
+                    throw $exception;
+                }
+
+                Tlog::getInstance()->error($exception->getMessage());
```
roadster31: Maybe you can log the exception before throwing it

local/modules/TheliaSmarty/Template/Plugins/Hook.php
```diff
@@ -414,9 +423,17 @@ protected function checkEmptyHook($params)
...
-            throw new \InvalidArgumentException(
+            $exception = new \InvalidArgumentException(
                 $this->translator->trans("Related hook name '%name' is not defined.", ['%name' => $hookName])
             );
+
+            if ($this->debug) {
+                throw $exception;
+            }
+
+            Tlog::getInstance()->error($exception->getMessage());
```
roadster31: And here too.

PR #2578, opened 2018-11-17
Improved the way the error message is passed to order failed page

roadster31:
This PR changes the way the error message is passed to the order failed page. This message was passed as a part of the URL, which may cause 404 if the message contains 'dot' or URL-significant characters.

The message is now passed as the "message" URL parameter. The previous method is preserved to prevent potential BC.

### DIFF
--- a/core/lib/Thelia/Module/BasePaymentModuleController.php
+++ b/core/lib/Thelia/Module/BasePaymentModuleController.php
@@ -252,10 +252,10 @@ abstract class BasePaymentModuleController extends BaseFrontController
     /**
      * Redirect the customer to the failure payment page. if $message is null, a generic message is displayed.
      *
-     * @param int         $order_id the order ID
+     * @param int         $orderId the order ID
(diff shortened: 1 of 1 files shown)


core/lib/Thelia/Module/BasePaymentModuleController.php
```diff
@@ -252,21 +252,18 @@ public function redirectToSuccessPage($orderId)
     /**
      * Redirect the customer to the failure payment page. if $message is null, a generic message is displayed.
      *
-     * @param int $orderId the order ID
-     * @param string|null $message an error message.
+     * @param int         $order_id the order ID
+     * @param string|null $message  an error message.
      */
-    public function redirectToFailurePage($orderId, $message)
+    public function redirectToFailurePage($order_id, $message)
```
lopes-vincent: I think this is better to use camel case for all variables names

lopes-vincent:
There are still some bad case for `$order_id`  in your PR

PR #2611, opened 2019-01-14
Fixed loop arguments cache initialisation

roadster31:
This PR fixes the loop argument cache initialization, to allow the loop argument extension system to work properly.


gillesbourgeat:
Thanks for this correction ! :+1:

PR #2615, opened 2019-01-23
feature_values filter is now working in Product loop

roadster31:
This PF fixes the `feature_value` filter of the Product loop. This filter is currently not working, as it operates on the `FREE_TEXT_VALUE` columns of the `FeatureProduct` table,  while free text feature values are now stored in the `FeatureAvI18n` table.

Additionally, the PR allows spaces ans special (, : ( ) | & ) characters in the provided values, with a standard escaping system. Something like `1: foo & bar | (fooo &baar), 4: *, 67: (foo\,fooo & foooo & baaar), 11:(abc & def\&ghi\|ttt)` is now allowed and is propely parsed.

The static methods `escape` and `unescape` are added to `IntToCombinedStringsListType` to provide an easy way to escape values.

### DIFF
--- a/core/lib/Thelia/Core/Template/Loop/Product.php
+++ b/core/lib/Thelia/Core/Template/Loop/Product.php
@@ -28,6 +28,7 @@ use Thelia\Model\CategoryQuery;
 use Thelia\Model\ConfigQuery;
 use Thelia\Model\CurrencyQuery;
 use Thelia\Model\Currency as CurrencyModel;
+use Thelia\Model\Map\FeatureAvI18nTableMap;
 use Thelia\Model\Map\ProductPriceTableMap;
 use Thelia\Model\Map\ProductSaleElementsTableMap;
 use Thelia\Model\Map\ProductTableMap;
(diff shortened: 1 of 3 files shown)


Alberttyty:
Hi, thanks for this fix but i found a problem :
if the feature value is an integer, it can replaces aliases already injects instead values.
Example : feature value expression == (1|2|3|4)

My contribution in **Core\Template\Loop\Product.php::manageFeatureValue**

```
foreach ($aliasMatches as $value => $alias) {
    // Do not replace aliases but only values in the feature value expression
    // Replace only the first value iteration
    if($previousReplaceLength) {

        $sqlWhereString = substr($sqlWhereString, 0, $previousReplaceLength)
                . preg_replace('/'.$value.'/', 'NOT ISNULL(`'.$alias.'`.ID)', substr($sqlWhereString, $previousReplaceLength), 1)
         ;

        $previousReplaceLength += strlen('NOT ISNULL(`'.$alias.'`.ID)');

     } else {
        $sqlWhereString = preg_replace('/'.$value.'/', 'NOT ISNULL(`'.$alias.'`.ID)', $sqlWhereString, 1);
        $previousReplaceLength = strlen('(NOT ISNULL(`'.$alias.'`.ID)');
    }

 }
```

Example : feature value expression == (1|2|3|4)

roadster31:
Thanks for this fix. Could you please make a PR ?

PR #2622, opened 2019-01-30
Propel schema generation is now protected from concurrency

roadster31:
This PR uses the Lock Symfony component to protect the Propel schema generation from concurrent requests.


gillesbourgeat:
Good, Thanks !

PR #2638, opened 2019-02-19
Remove Tlog in Propel init

lopes-vincent:
Can't use Tlog here because if Propel fail to init the ConfigQuery class is not generated.
And Tlog need ConfigQuery to be initialized.

### DIFF
--- a/core/lib/Thelia/Core/PropelInitService.php
+++ b/core/lib/Thelia/Core/PropelInitService.php
@@ -389,8 +389,6 @@ class PropelInitService
                 $theliaDatabaseConnection->useDebug(true);
             }
         } catch (\Throwable $th) {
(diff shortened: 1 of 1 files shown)


core/lib/Thelia/Core/PropelInitService.php
```diff
@@ -389,8 +389,6 @@ public function init($force = false)
                 $theliaDatabaseConnection->useDebug(true);
             }
         } catch (\Throwable $th) {
```
roadster31: The catch clause is not needed. The finally clause alone is enough.

gillesbourgeat: maybe the cache should be reset at this time to avoid hidden errors at next boot

PR #2649, opened 2019-03-02
Docker & Docker compose update

nicolas-zozol:
There is one bug in Dockerfile dependencies, and the docker-compose.yml was not perfect


roadster31:
Thanks :)

PR #2651, opened 2019-03-06
Allow to load tax rule without country

lopes-vincent:
With this PR you can load a tax rule without country in Calculator

### DIFF
--- a/core/lib/Thelia/TaxEngine/Calculator.php
+++ b/core/lib/Thelia/TaxEngine/Calculator.php
@@ -97,6 +97,25 @@ class Calculator
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


core/lib/Thelia/TaxEngine/Calculator.php
```diff
@@ -97,6 +97,25 @@ public function loadTaxRule(TaxRule $taxRule, Country $country, Product $product
         return $this;
     }
```
gillesbourgeat: Hello,
Missing phpdoc `@since 2.4`

lopes-vincent: Done

PR #2652, opened 2019-03-14
More than one zip code template for a country

roadster31:
This PR allow more than one zip code format for a country. Several zip format separated bu commas could be entered for each country.


roadster31:
Check tpl generation, addtionnal languages shoud not be there.

PR #2677, opened 2019-06-26
Profile management improvement

roadster31:
This PR improves the administrator profiles edition, which was hard to manage. A global toggle button is and a line toggle button are added.
The resources are now sorted by code.
A new "order" parameter is added to the Resources loop.
A few improvements to the UI (standardization of module order in various select menus, direct access to the module configuration by clicking on the module title).


templates/frontOffice/default/includes/add-to-cart-error.html
```diff
@@ -0,0 +1,21 @@
+{* This page should not replace the current previous URL *}
```
lopes-vincent: I'm not sure this change need to be in this PR

templates/frontOffice/default/product.html
```diff
@@ -230,11 +230,16 @@ <h1 class="name"><span itemprop="name" content="{$TITLE}">{$TITLE}</span><span i
                     {form_field field="append"}
                         <input type="hidden" name="{$name}" value="1">
                     {/form_field}
+
                     {form_field form=$form field="success_url"}
                         <input type="hidden" name="{$name}" value="{url path='/cart'}">
                     {/form_field}
 
-                    {if $form_error}<div class="alert alert-error">{$form_error_message}</div>{/if}
+                    {form_field form=$form field="error_url"}
```
lopes-vincent: Same here, I'm not sure this change need to be in this PR

templates/frontOffice/default/includes/add-to-cart-error.html
roadster31: File has been removed from PR

templates/frontOffice/default/product.html
roadster31: Original product.html has been restored

PR #2685, opened 2019-07-04
Added arrow navigation to documents and images management

NicolasBarbey:


templates/backOffice/default/document-edit.html
```diff
@@ -9,7 +9,7 @@
 
         <div id="wrapper" class="container">
 
-            {loop type="document" name="document_edit" source="{$documentType}" id="{$documentId}" backend_context="1" lang="$edit_language_id" visible="*"}
+            {loop type="document" name="document_edit" source="{$documentType}" id="{$documentId}" backend_context="1" lang="$edit_language_id" visible="*" with_prev_next_info="true"}
```
roadster31: You can write `true` instead of `"true"`, and also `backend_context=1` instead of `backend_context="1"`

Quotes are required for string literals, but not for numbers, boolean values or Smarty variables.

PR #2688, opened 2019-09-04
Fix a bug on the images tab

NicolasBarbey:


roadster31:
Which bug ?

lopes-vincent:
Fixed by #2715

PR #2723, opened 2020-03-06
Fixed OrderProduct loop and Order model non rounded prices

Lurivar:
Round the prices returned by the Order and Order Product Loops to a XX,XX format to prevent prices being the round of product prices instead of the sum of rounded prices.


lopes-vincent:
I think this is not a good idea to change the old order recorded with bad rounded amount

lopes-vincent:
Fixed by #2735

PR #2725, opened 2020-03-13
Better default tax rules names

Lurivar:
Better default tax rules names which should be less confusing to people not aware of how tax rules work.


lopes-vincent:
Replaced by #2771

PR #2735, opened 2020-03-27
Tax calculation fixes, revamped

roadster31:
This is a subset of #2680

This PR implements the "sum of rounded taxes" method for tax calculations, to prevent total price calculation errors when large quantities are ordered.

The PR also introduces an improved order-invoice order total block, and a PDF invoice improvement.


core/lib/Thelia/Core/Template/Loop/Order.php
```diff
@@ -320,13 +321,21 @@ public function buildModelCriteria()
      */
     public function parseResults(LoopResult $loopResult)
     {
+        $lastLegacyOrderId = ConfigQuery::read('last_legacy_order_id', 0);
```
lopes-vincent: The term legacy is not really clear maybe last_miscalculated_order_id or last_not_rounded_order_id ?

roadster31: `last_old_style_rounding_order_id` ?

PR #2736, opened 2020-03-27
Optimized exports with JSON cache file and SQL request

Cav0n:


core/lib/Thelia/ImportExport/Export/JsonFileAbstractExport.php
```diff
@@ -47,9 +55,9 @@ public function rewind()
 
             // Check if $data is a path to a json file
             if (\is_string($data)
-                && ".json" === substr($data, -5)
-                && file_exists($data))
-            {
+                && '.json' === substr($data, -5)
```
lopes-vincent: This change is not really useful 😜

PR #2737, opened 2020-03-27
Add php < 7.4 requirements

lopes-vincent:


core/lib/Thelia/Install/CheckPermission.php
```diff
@@ -50,6 +50,11 @@ class CheckPermission extends BaseInstall
         'upload_max_filesize' => 2097152
     );
 
+    protected $phpExpectedVerions = array(
+        'min' => '5.5',
+        'max' => '7.4'
```
roadster31: Maybe putting 7.3 here, and checking max version with > instead of >= would be more clear to the reader

lopes-vincent: I also think php ! >= 7.4 is not really clear but if I set ! > 7.3 the test will faill on 7.3.X like 7.3.16

PR #2788, opened 2020-10-15
fix issue #2787 TinyMce add link to img doesn't work

zzuutt:
fix issue #2787 
on ne peut pas ajouter un lien sur une image
Uncaught DOMException: Index or size is negative or greater than the allowed amount

il faut remplacer la ligne 17792 dans le fichier tinymce.js (non minified)

```
if (node && node.tagName == 'IMG') {
  self.getSel().setBaseAndExtent(node, 0, node, 1);
}
```

par

```
if (node && node.tagName == 'IMG') {
  self.getSel().setBaseAndExtent(node, 0, node, 0);
}
```


lopes-vincent:
Normally you shouldn't modify this file as it is a lib file and your modification will be erased if we update TinyMCE.
But we are not planning to update TInyMCE as we are checking for a better alternative to replace it. So it's ok for this time.

PR #2789, opened 2020-10-22
fix State parameter

TenseiV:


lopes-vincent:
Replace by #2790 because modify DelliveryModuleInterface will break existing module

PR #2799, opened 2020-11-22
Deep cloning of template definition

roadster31:
The template definition, is not deeply cloned, which may cause unexpected side effects on template search when using `template` parameter of `stylesheet`, `javascript` and `image` Smarty functions, like this :

`{stylesheet file='/assets/css/fontawesome/font-awesome.less' filters='less' source='Tinymce' template='my_template'}`

### DIFF
--- a/local/modules/TheliaSmarty/Template/SmartyParser.php
+++ b/local/modules/TheliaSmarty/Template/SmartyParser.php
@@ -416,7 +416,7 @@ class SmartyParser extends Smarty implements ParserInterface
      */
     public function getTemplateDefinition($webAssetTemplateName = false)
     {
-        // Deep cloen of template definition. We could change the template descriptor of template definition,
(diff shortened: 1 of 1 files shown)


local/modules/TheliaSmarty/Template/SmartyParser.php
```diff
@@ -416,7 +416,11 @@ public function setTemplateDefinition(TemplateDefinition $templateDefinition, $f
      */
     public function getTemplateDefinition($webAssetTemplateName = false)
     {
-        $ret = clone $this->templateDefinition;
+        // Deep cloen of template definition. We could change the template descriptor of template definition,
```
InformatiqueProg, line 419: deep clone  ? no 😉

lopes-vincent: yes thanks

PR #2801, opened 2020-12-02
add a command to switch the front template from CLI

Lucanis:


core/lib/Thelia/Command/FrontTemplate.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $this
+        ->setName("template:front")
+        ->setDescription("set front template")
+        ->addArgument(
+            "template",
+            InputArgument::REQUIRED,
+            "template to activate"
+        );
+        ;
```
gillesbourgeat, line 30:
```suggestion
```

roadster31:
Why don't you just make a command which changes the value of a configuration variables, such as `config:set <var> <val>` ?

PR #2805, opened 2020-12-24
Add i18n for HookAnalytics configurations

NicolasBarbey:


local/modules/HookAnalytics/Form/Configuration.php
```diff
@@ -27,7 +27,9 @@ protected function buildForm()
     {
         $form = $this->formBuilder;
 
-        $value = ConfigQuery::read("hookanalytics_trackingcode", "");
+        $lang = $this->getRequest()->getSession()->get("thelia.admin.edition.lang");
+
+        $value = HookAnalytics::getConfigValue("hookanalytics_trackingcode", "", $lang->getLocale());
```
lopes-vincent, line 32: Please can you verify that the lang is not null before calling his method ? 
And if null maybe take the default lang.

local/modules/HookAnalytics/HookAnalytics.php
```diff
@@ -12,9 +12,21 @@
...
+use Thelia\Model\ConfigQuery;
+use Thelia\Model\LangQuery;
 use Thelia\Module\BaseModule;
 
 class HookAnalytics extends BaseModule
 {
-
+    public function update($currentVersion, $newVersion, ConnectionInterface $con = null)
+    {
+        if (version_compare($newVersion, $currentVersion, ">")){
```
lopes-vincent, line 24: This will copy the trackingcode at each update even for all future updates and erase filed values. You should check that the current version is less or equal to 2.4.3 and the new version greater than 2.4.3

local/modules/HookAnalytics/HookAnalytics.php
```diff
@@ -12,9 +12,21 @@
...
+            $langs = LangQuery::create()->filterByActive()->find();
+            $config = ConfigQuery::read("hookanalytics_trackingcode", "");
+            foreach ($langs as $lang){
+                self::setConfigValue('hookanalytics_trackingcode', $config, $lang->getLocale());
```
lopes-vincent, line 28: Please verify "hookanalytics_trackingcode" for the lang is not null before erasing it.

lopes-vincent (changes requested):

PR #2814, opened 2021-02-03
Fix export end date calculation

roadster31:
The export end date calculation was wrong, and gave wrong results (e.g. 2021-01-28 instead of 2021-01-31).


roadster31:
Merci m'sieur 🌙

PR #2835, opened 2021-04-19
Fix init template modern

crobert:
- BestSellers is used in default modern template, so it needs to be in default composer and to be activated on initial setup
- Adding some minors template fix


lopes-vincent:
There are no errors if BestSellers is not installed or not active, so the module is not mandatory

Issue #2829, opened 2021-03-15
Contact name is lost when contact form is submited


roadster31:
The `$name` attribute of ContactEvent is conflicting with  `$name` Symfony\Component\EventDispatcher\Event attribute

The name entered in the contact form is correctly set in ContactEvent constructor, but is replaced by the event name (`thelia.contact.submit`) when the event is dispatched here : https://github.com/thelia/thelia/blob/48ebc1644e0a81005310e5d9b06cf8e94e7e86fe/local/modules/Front/Controller/ContactController.php#L53

Patatouix:
I confirm the bug. 
Moreover i have an issue with ContactEvent : I added some extra fields to the contact form and unfortunately i cannot access them from the event, because there is no getter for `$this->form`.

anoziere:
This was true for the legacy symfony/event-dispatcher Event class (pre-4.3) which carried a mutable $name set by the dispatcher, but ActionEvent now extends the modern Symfony\Contracts\EventDispatcher\Event (no $name property), and EventDispatcher::dispatch() never mutates the event. Verified the name survives dispatch in both Thelia 2.6 and Thelia 3. Closing as obsolete/fixed by the Symfony upgrade. Note: the secondary ask in the comments (a getForm() getter on ContactEvent) is still genuinely missing in both branches — worth a separate small enhancement issue if still wanted.

PR #2839, opened 2021-05-05
FIX conflict between ContactEvent->name and Event->name

Patatouix:
#2829 Changed the name of `ContactForm `field `name `to `fullname `in order to avoid a conflict between the `$name` property of `ContactEvent ` and the `$name` property of its parent `Event`.


lopes-vincent:
Sorry but your PR doesn't contain any compatibility with older modules or website using Thelia. If they use the "name" field it will break.

PR #2840, opened 2021-05-05
Enhancement of ContactEvent when ContactForm fields are modified

Patatouix:
-  If some of the basic fields (name/email/subject/message) are removed from `ContactForm `(for example with `FORM_AFTER_BUILD `event), prevents php error from `$form->get('name')` in `ContactEvent`
-  If some custom fields are added to `ContactForm `, allows to retrieve data from these fields in `ContactEvent `(which is not possible in the current code)

### DIFF
--- a/core/lib/Thelia/Core/Event/Contact/ContactEvent.php
+++ b/core/lib/Thelia/Core/Event/Contact/ContactEvent.php
@@ -41,11 +41,14 @@ class ContactEvent extends ActionEvent
     public function __construct(Form $form)
     {
         $this->form = $form;
+    }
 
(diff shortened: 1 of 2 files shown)


lopes-vincent:
Same as precedent PR this totally break the compatibility.
Thelia contact page is really basic. And as each site will want a different contact site with more or less complexity (sometimes it won't be just to add more field) we can't really do better.
If you want to do more I suggest you to do your own form with your own controller, it's really easier than adding field to the Thelia contact form.

But thank you anyway for trying to help us improve Thelia ! ;)

PR #2986, opened 2022-05-30
feat: config for prettier and linters started + husky and lint-staged…

anonymze:
feat: added .eslintcache in gitignore


pre-commit
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+## add paths to directories where you want to lint your assets
+
+### front template
+cd templates/frontOffice/modern && npx lint-staged
```
Lucanis, line 7: if [[ $(git diff --name-only HEAD ./templates/frontOffice/modern) != "" ]]; then
    echo "linting modern frontOffice template"
    cd templates/frontOffice/modern && npx lint-staged
fi

PR #3033, opened 2022-09-01
Fixed front/back template handling in webpack_encore.yaml

roadster31:
The front template could only be defined in a `.env` file (not in the BO) because `config/packages/webpack_encore.yaml` uses only `ACTIVE_FRONT_TEMPLATE` and `ACTIVE_ADMIN_TEMPLATE` variables.

This PR fixes the problem, by defining two container parameters `thelia_front_template` and `thelia_backoffice_template`, that are now used in `config/packages/webpack_encore.yaml`  instead of the variables defined in `.env` files.

Additionally, the PR fixes the autoloading of classes defined in `components` directories of the back and front templates, where only `ACTIVE_FRONT_TEMPLATE` and `ACTIVE_ADMIN_TEMPLATE` variables were used, just as explained above.

### DIFF
--- a/core/lib/Thelia/Config/Resources/config.xml
+++ b/core/lib/Thelia/Config/Resources/config.xml
@@ -95,7 +95,9 @@
         <service id="Psr\Container\ContainerInterface" alias="service_container"/>
 
         <!-- Thelia template helper -->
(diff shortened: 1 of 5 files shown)


lopes-vincent:
#3039

PR #3107, opened 2023-07-07
Change i18n "My Profile" to "My profile"

LoicMo43:


templates/frontOffice/modern/I18n/fr_FR.php
```diff
@@ -116,7 +116,7 @@
   'Max price' => 'Prix max',
   'Min price' => 'Prix min',
   'My Addresses' => 'Mes adresses',
-  'My Profile' => 'Mon Profil',
+  'My profile' => 'Mon profil',
```
lopes-vincent, line 119: If you change the source phrase ("My Profile" in English) you must also change it in the template {intl}

PR #3176, opened 2024-02-27
add fields lastname and firstname for default adress into customer edition page

ajouguelet:


Elyos59:
It should be done by module not by core modification...

PR #3211, opened 2024-05-02
Update ProductController.php

Elyos59:
Free text value could be equal to 0.


Elyos59:
Other files must be modified...

Elyos59:
See https://github.com/thelia/thelia/pull/3212

PR #3221, opened 2024-06-18
Feat/payment modules

pierrebaguet:


templates/frontOffice/modern/components/React/Checkout/hooks/useValidPaymentModules.ts
```diff
@@ -4,12 +4,10 @@ import { usePaymentModulessQuery } from '@openstudio/thelia-api-utils';
 export default function useValidPaymentModules(type?: any) {
```
leanormandon, line 4: type useless

pierrebaguet: Corrigé

leanormandon:
Approved

Issue #3311, opened 2025-04-28
REAL_PRICE_TAX empty in OrderProduct Loop


Elyos59:
### Thelia version(s) affected

2.5.4

### Description

Hi!

When no rule tax found for a country no order_product_tax is saved in database.

So, OrderProduct Loop returns an empty REAL_PRICE_TAX and Thelia returns an error in order-edit.html template :

`The following error message has been found: Unsupported operand types: string * float`

### How to reproduce

Read description.

### Possible Solution

- Returns 0 if REAL_PRICE_TAX is empty
OR
- Get TVA RULE set by default when order_product are saved in database <-- Best solution for me

PR #3314, opened 2025-04-30
Update OrderProduct.php

Elyos59:
Prevent empty value when no tax rule.

https://github.com/thelia/thelia/issues/3311


roadster31:
👍

PR #3321, opened 2025-05-16
Feat/use public dir instead of web

Lucanis:


gillesbourgeat:
Hello Damien, and thank you for your contribution.

You are deleting the file web/.htaccess, but it would be better to move it to public/.htaccess.
The same change should also be made in the repository https://github.com/thelia/thelia-project.

best regards,

PR #3328, opened 2025-06-26
fix : circular reference, NotInFilter with multiple values, force join

ArthurLashermes:


ArthurLashermes:
Technical Summary of the Changes:
Handling Circular Dependencies:
The modelToResource logic did not take serialization groups into account, which could lead to circular dependencies when using recursion. To avoid this, an excludedGroups parameter was introduced to block specific groups and prevent infinite loops.

Optimizing Automatic Joins with Propel:
By default, joins are automatically added for simple property types. However, for properties of type array, automatic joins are skipped — unless used in the context of a filter or sort.
This behavior helps reduce the total number of joins in a query, since MariaDB has a known limitation of around 60 joins per query. Avoiding unnecessary joins helps prevent hitting that limit and triggering errors.

Ability to Force a Join When Needed:
In some specific cases, such as when applying advanced filters, it becomes necessary to force a join even for array properties. A mechanism was added to explicitly allow forcing a join in such scenarios, since the previous system based on filter names was not sufficient or reliable in all cases.

Fix for NotInFilter:
The NotInFilter was not properly handling multiple values. Although Swagger allowed passing multiple values to exclude, only the last value was actually taken into account. This has been fixed so that all provided values are now excluded correctly.

Issue #2562, opened 2018-11-11
Custom order status should declare which of the "canonical" status are covered


roadster31:
When a new order status is defined in the back-office, it should declare which of the "canonical" status are covered.

Related to #1707

anoziere:
Design settled after review (including community PR #3331, which names the right requirement but whose branch predates the Symfony 7.4 / API Platform 4 rewrite and cannot be rebased): a nullable `order_status.equivalent_code` rather than per-status booleans — the six canonical codes are mutually exclusive states of one order, so one equivalence value per custom status matches the domain, and `isPaid()`-style helpers resolve through an effective code (`equivalent_code ?: code`). Fully backward compatible: existing custom statuses with no equivalence behave exactly as today. The back office gains an optional 'Equivalent to' select limited to the six canonical codes, on non-protected statuses only.

Implementation lands on `main` (Thelia 3) first, where the schema is still cheap to change; a 2.6 backport can follow if there is demand.

PR #3331, opened 2025-07-19
Allow to define custom status as paid status

Elyos59:
For example, adding the status "Validated" for certain customers who paid on delivery. But to manage inventory effectively or if a script deletes unpaid orders after X days, the order must be considered "paid".


anoziere:
Hey, thanks for the PR. A few things before we can move forward:

- Could you add a short description of the use case? What kind of custom status are you working with?
- The commit history needs a squash , 9 commits with "Update OrderStatus.php" x3 is hard to review
- There's no migration SQL for existing installs — modifying `insert.sql` and `thelia.sql` alone won't upgrade running instances

Also heads up: the back-office is being rewritten for Thelia 3. If this is still relevant for 2.x maintenance, let's get it in shape. Otherwise we might close this one.

Elyos59:
> * The commit history needs a squash , 9 commits with "Update OrderStatus.php" x3 is hard to review

I don't know how to do that

> * There's no migration SQL for existing installs — modifying `insert.sql` and `thelia.sql` alone won't upgrade running instances

What then?

anoziere:
Thanks for this — the requirement is right and it drove the design now landing for Thelia 3 (see https://github.com/thelia/thelia/issues/2562: nullable `order_status.equivalent_code` + an 'Equivalent to' select in the back office). This branch itself forked before the Symfony 7.4 / API Platform 4 rewrite and can no longer be rebased (the diff exceeds GitHub's display limits and conflicts throughout), so it will be closed once the new implementation is merged. Your review on that implementation would be very welcome.

anoziere:
The feature you asked for is now on `main`: #3533 adds a nullable `order_status.equivalent_code` column plus an "Equivalent to" select in the back office, so a custom status can declare the canonical status it stands for (paid, sent, canceled). Stock handling and any script that looks for paid orders read the equivalent code, which is what you needed for pay-on-delivery.

Closing this branch: it forked before the Symfony 7.4 and API Platform 4 rewrite, so it cannot be rebased any more (the diff exceeds GitHub's display limits and conflicts throughout). Thanks for pushing the use case, it shaped the final design. Reviews on #3533 are welcome.

PR #3436, opened 2026-07-20
Remove maximumNumberOfProcesses limitation to keep analysis fast on modern computers.

staabm:
In CI PHPStan will automatically the maximum number of CPUs to use


anoziere:
Thanks for this. Your branch could no longer be rebased: it was opened before `main` became the Thelia 3 line, so its diff now spans the whole rewrite. The change itself is in, unchanged, in #3537 on `main` and #3538 on `2.6`, with the reasoning credited to you. Closing this one.

PR #3532, opened 2026-08-11
feat: store the customer discount rate on the order

### LINKED ISSUE #2631: Store customer discount in order information
At the moment, this discount is not saved, and can't be displayed safely on invoices.

anoziere:
A customer discount is baked into the cart item prices, so once the customer's
rate changes there is no way to know which rate produced a past order. The
existing `order.discount` column already means "coupon total", so it cannot be
reused for this.

- new `order.customer_discount_rate` column (`schema.xml` + `setup/thelia.sql`,
  `order` and `order_version`), holding the rate as a percentage
- `OrderFactory` copies the customer's rate in force when the order is created
- `order` loop exposes `CUSTOMER_DISCOUNT_RATE`; read-only `customerDiscountRate`
  on the `Order` API resource (admin and front single read)
- integration test: an order placed at 10% still reads 10% after the customer's
  rate is raised to 25%

Migration: fresh installs get the column from `setup/thelia.sql`. Existing
installs need an `ALTER TABLE` on `order` and `order_version`, to be folded into
the next `setup/update/sql/` release file rather than guessed here.

Invoice display lives in `thelia-templates/pdf`, a separate repository: see the
companion PR https://github.com/thelia-templates/pdf/pull/7. Merge this one first,
the template reads the new loop output.
No front-office display is added — Flexy shows no discount at all today, so that
is separate UI scope.

Fixes #2631


anoziere:
CI is red for a reason outside this branch.

`thelia/core` requires `thelia/config`, and `composer/installers` extracts that
package over `local/config/` (`type: thelia-local`, `installer-name: config`).
So the `composer install` step replaces the patched `local/config/schema.xml`
with the published `thelia/config` 3.0.0-beta1 content before `bin/test-prepare`
generates the Propel models:

```
- Installing thelia/config (3.0.0-beta1): Extracting archive
...
Propel\Runtime\Exception\BadMethodCallException: Call to undefined method: setCustomerDiscountRate.
var/propel/test/model/Thelia/Model/Base/Order.php:4875
```

`setup/thelia.sql` is not affected (the repository copy wins), so the test
database does have the column — only the generated model lacks it. cs-fixer,
PHPStan and the unit suite are green; the 4 integration errors and the 2
Mail/Pdf render failures all come from that single missing setter.

Any new column on a core table hits this. Unblocking it needs `local/config` to
be dispatched to `thelia/config` and tagged; the branch itself has nothing left
to fix.

Issue #2866, opened 2021-09-09
Add ISO 4217 numeric currency code to currency table


roadster31:
The currency table currently contains the alphanumeric version of the ISO 4217 currency code (e.g. EUR), but not the numeric version (e.g. 978) which is often used by payment systems.

See https://en.wikipedia.org/wiki/ISO_4217#Numeric_codes for details.

anoziere:
Done in #3524, merged into the `2.6` branch: the `currency` table gains an `isocode_numeric` column (ISO 4217 numeric code), seeded for all bundled currencies on fresh installs and backfilled by the 2.6.2 update script on existing shops, exposed in the currency loop and on the API resource. Verified against both the fresh-install and the upgrade paths on a real MariaDB before merge. Closing manually: `Fixes` keywords only auto-close from the default branch (`main`).

PR #3549, opened 2026-08-11
feat: add the ISO 4217 numeric currency code

### LINKED ISSUE #3535: Port the ISO 4217 numeric currency code to Thelia 3
Counterpart of #2866 (fixed on `2.6` by https://github.com/thelia/thelia/pull/3524): add `currency.isocode_numeric` to `main` (schema.xml + DDL + seed values + loop/API exposure).

Heads-up for the implementer: any `local/config/schema.xml` change on `main` fails CI until `local/config` is re-dispatched to `thelia/config` and tagged (see the packaging trap documented on #3532/#3533). Batch this with the next schema-touching change to share one `thelia/config` release.

anoziere:
Thelia 3 counterpart of #2866, ported from the 2.6 fix (#3524): the `currency` table had no ISO 4217 numeric code, which several payment and accounting integrations require alongside the alphabetic code.

Adds the `currency.isocode_numeric` column (`VARCHAR(3)`, nullable) in `local/config/schema.xml` and `setup/thelia.sql`, seeds the 48 shipped currencies with the same numeric codes as 2.6 in `setup/insert.sql` and `setup/insert.sql.tpl` (leading zeros preserved, e.g. ARS `032`, BOB `068`), and exposes the value through `CurrencyCreateEvent`, `Action\Currency`, `CurrencyFacade` + its create/update DTOs, the `currency` loop (`ISOCODE_NUMERIC`) and the `Currency` API resource. No BC break: the column is nullable and every new argument is optional and defaults to `null`.

No `setup/update/sql` script is included: the latest one is `3.0.0-alpha1.sql` and there is none for the current `3.0.0-beta1`, so the column should be batched into the next version script together with `order.customer_discount_rate` from #3532, which is in the same situation.

The back-office form field lives in the `default-twig` theme repository and will be added there separately.

**Expected CI failure until `thelia/config` is re-dispatched and re-tagged**: the CI `composer install` extracts `thelia/config` over `local/config/`, replacing this branch's `schema.xml` before Propel generates the models, so the new column is missing from `Thelia\Model\Currency`. Expected errors, all from that single cause:
- PHPStan: `Call to an undefined method Thelia\Model\Currency::setIsocodeNumeric()` in `core/lib/Thelia/Action/Currency.php` (2 occurrences) and `Call to an undefined method Thelia\Model\Currency::getIsocodeNumeric()` in `core/lib/Thelia/Core/Template/Loop/Currency.php` (1 occurrence).
- PHPUnit integration suite: `CurrencyActionTest::testCreateCurrencyPersistsWithUppercaseCode` and `CurrencyActionTest::testUpdateCurrencyRate`, same undefined methods.

Everything else is green: `php-cs-fixer` reports 0 of 9 files to fix, the seed block loads into MariaDB (48 rows, all codes exactly 3 characters), and the unit suite `CurrencyFacadeTest` passes (7 tests, 60 assertions) — verified red before the change, green after.

Fixes #3535


anoziere:
Merging with the CI red, per the packaging cycle documented on #3532 and #3533: the CI `composer install` extracts `thelia/config` into `local/config/`, which overwrites the branch `schema.xml` before Propel generates the models, so a schema change cannot be green until `local/config` is dispatched and `thelia/config` is tagged again.

Every failure on this run is that overwrite: five `Call to undefined method: (get|set)IsocodeNumeric` errors, plus `MailRenderTest::testRendersOrderConfirmationForAnOrder`, which renders a template reading the new loop output. cs-diff, PHPStan and the four other suites are green.

To verify after the `thelia/config` release: re-run this workflow on `main` and check that both the currency errors and the mail render failure are gone. If the mail render one survives, it is not this change and needs its own look.

Issue #3073, opened 2022-12-22
Hook prioritize


mabruchet:
### Description

We have issue with hooks.

When you have a module hook that depends on another hook of another module, the position influences the insertion of the latter. And so when you do a `php Thelia hook:clean` it breaks the order of the insertion because it will take the order of the modules to recreate the hooks. 

Would it be possible to add a priority on the hooks as for the events ? 
  
Solution talk with  @lopes-vincent 

### Example

- Don't work ( depends on same hook thelia.blocks.plugins )
  - ProjectModule 
  - TheliaLibrary
  
- Work ( depends on same hook thelia.blocks.plugins )
  - TheliaLibrary
  - ProjectModule

anoziere:
Mapped the mechanism on both lines before answering, because the report is accurate and the cause is not where one would expect.

There is no runtime ordering query. The order of a hook's listeners is frozen at container compile time into Symfony listener priorities: `ModuleHookQuery::create()->orderByHookId()->orderByPosition()->orderById()->find()` in `RegisterHookListenersPass::addHooksMethodCall()` (2.6 core/lib/Thelia/Core/DependencyInjection/Compiler/RegisterHookListenersPass.php:247-251, `main` :203-207), turned into `ModuleHook::MAX_POSITION - $moduleHook->getPosition()` (2.6 :299-309, `main` :255-265). `module.position` and `hook.position` are read nowhere in that path. So `module_hook.position` is the only lever, and it is settable only through the back office.

`hook:clean` truncates that table. `HookCleanCommand::deleteHooks()` (2.6 core/lib/Thelia/Command/HookCleanCommand.php:126-145, `main` :127-148) calls `ModuleHookQuery::create()->deleteAll()` and `IgnoredModuleHookQuery::create()->deleteAll()`. On the next container build every row is recreated with the `MAX_POSITION` sentinel in module load order (module.position), the sort collapses to insertion order and the renumbering loop rewrites 1..N. Two consequences, not one: every ordering the administrator set is lost, and every hook that had been explicitly ignored comes back.

Nothing declarative exists today. The `hook.event_listener` tag accepts only name, event, method, type, active and templates (thelia-1.0.xsd:206-213), and `getSubscribedHooks()` only type, method and template. A module cannot state its order at all.

Three options, in increasing cost.

**A. Make hook:clean preserve the ordering it does not need to reset** (2.6 and main, cost small, risk low)

[...]

PR #3561, opened 2026-08-11
fix: keep the hook positions and the ignored hooks when running hook:clean

anoziere:
`php Thelia hook:clean` truncates `module_hook` and `ignored_module_hook` (`HookCleanCommand::deleteHooks()`, core/lib/Thelia/Command/HookCleanCommand.php:127). `module_hook.position` is the only lever that orders the listeners of a hook: `RegisterHookListenersPass::addHooksMethodCall()` reads it (RegisterHookListenersPass.php:205) and turns it into the Symfony listener priority (`ModuleHook::MAX_POSITION - $moduleHook->getPosition()`, :263). A maintenance command therefore resets every rendering order the administrator set in the back office, and brings back every hook that was explicitly removed there.

The command now records the `(module_id, hook_id, method) -> position, active` map before deleting the rows, and `RegisterHookListenersPass::registerHook()` reapplies it when it recreates them during the next container build. The map is kept in the `config` table (`saved-module-hook-configuration`, hidden and secured): the command wipes the cache directory right after the deletion, and the container is rebuilt by another process, possibly under another system user. It is deleted once the hooks have been recreated. `ignored_module_hook` is no longer touched. No schema change.

The purpose of the command is unchanged: hooks are still deleted and recreated from the module declarations, so rows that are no longer declared are still purged.

`--reset-positions` restores the previous behaviour: drop the saved configuration and clear the ignored hooks.

`tests/Integration/Command/HookCleanCommandTest.php` covers the two halves, deletion and recreation, but not the container build itself: running the command through the console would clear the cache directory of the kernel running the tests. The whole path was checked by hand on an install, where a hook set to position 42 and inactive, plus an ignored hook, survive `hook:clean` followed by a container rebuild.

Refs #3073. The issue also asks for a position stated in the hook declaration itself, which needs a new `module_hook` column and a ruling on who wins between the module and the administrator, so it stays open.

### DIFF
--- a/core/lib/Thelia/Command/HookCleanCommand.php
+++ b/core/lib/Thelia/Command/HookCleanCommand.php
@@ -24,6 +24,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Thelia\Core\Event\Cache\CacheEvent;
 use Thelia\Core\Event\TheliaEvents;
+use Thelia\Core\Hook\ModuleHookConfigurationStore;
 use Thelia\Model\IgnoredModuleHookQuery;
 use Thelia\Model\Module;
 use Thelia\Model\ModuleHookQuery;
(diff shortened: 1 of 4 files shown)


anoziere:
Merging with the CI red: `main` is red at base right now, for the `thelia/config` overwrite documented on #3549. This run shows the same five `IsocodeNumeric` errors and the same `MailRenderTest` failure as the base branch, and the 16 tests this branch adds all pass (423 tests against 407 before). Nothing here is new.

Issue #3663, opened 2026-08-12
The Flexy front renders no download link for a virtual product


anoziere:
Found while mapping the virtual product storage for #3520.

The route exists — `customer.order.product.download`, declared by the Front module (`Config/front.xml:100-103`), which Flexy requires through `thelia/front-module`. But no Flexy template references it, and no template in the theme mentions `virtual` at all: `account-order.html.twig` lists the order lines without ever offering the file.

So on a standard Thelia 3 install, a customer who buys a virtual product **cannot reach their file** unless they type the URL by hand. The whole feature is wired end to end in the back office, the core and the delivery module, and stops at the last step.

The link belongs on the order detail page, on the lines whose product is virtual, and should only appear once the order reaches a status that authorises the download — the route already enforces that, so the template must not promise more than the route allows.

anoziere:
The API half of this is merged: https://github.com/thelia/thelia/pull/3677.

A theme builds the order page from a single call to `GET /front/account/orders/{id}`, and
`OrderProduct::$virtual` / `$virtualDocument` did not travel with it — they only carried
`front:order_product:read:single`, a group the order operation never activates. Writing
`{% if orderProduct.virtual %}` in a template would have been a silent no-op. Both properties
now also carry `front:order:read:single`, so the order payload states which lines are virtual
and which of them have a file behind them.

The front-office half — a download route and the link on the order page — is not in yet: the
Flexy theme is frozen while its Tailwind v4 rework lands, and patching it now would only create
conflicts. It follows once that merge is done, which is why this issue stays open.

PR #3677, opened 2026-08-13
fix: expose the virtual flags of an order product on the front order payload

anoziere:
`GET /front/account/orders/{id}` normalizes with the `front:order:read` and
`front:order:read:single` groups, and those propagate to the nested `orderProducts`
collection. `OrderProduct::$virtual` and `OrderProduct::$virtualDocument` only carried
`front:order_product:read:single`, a group activated by the standalone
`GET /front/account/order_products/{id}` operation and never present in the order context.

A theme that reads the order page from that single operation therefore had no way to tell
a virtual line from a physical one: `{% if orderProduct.virtual %}` silently evaluated to
null. Both properties now also carry `Order::GROUP_FRONT_READ_SINGLE`.

No new class of information is exposed: the neighbouring order product operation, secured
by the same ownership rule, already returned both fields to the same customer.

Refs #3663

### DIFF
--- a/core/lib/Thelia/Api/Resource/OrderProduct.php
+++ b/core/lib/Thelia/Api/Resource/OrderProduct.php
@@ -266,7 +266,13 @@ class OrderProduct implements PropelResourceInterface
     #[Groups([self::GROUP_ADMIN_READ, self::GROUP_ADMIN_WRITE, Order::GROUP_ADMIN_WRITE, self::GROUP_FRONT_READ_SINGLE])]
     public ?int $parent = null;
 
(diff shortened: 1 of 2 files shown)


anoziere:
The `PHPStan level 5 (BO Twig bundle)` job is red on this pull request for a reason it does not
introduce. It fails identically on `main` — `Parameter #1 $array (list<int>) of array_values is
already a list, call has no effect.` — at 7e821ea (the base of this branch, before this change)
and on every `main` push since #3671:

| run | head | conclusion |
|---|---|---|
| https://github.com/thelia/thelia/actions/runs/31613368851 | 7e821ea (base) | failure |
| https://github.com/thelia/thelia/actions/runs/31612904566 | 5fca222 (#3671) | failure |

This branch touches no file under `templates/backOffice/default-twig/`. The `CI (coding standard,
static analysis, tests)` job, which runs cs-diff, PHPStan on `core` and the five PHPUnit suites,
passes on 8.3 and 8.4.

PR #3684, opened 2026-08-13
fix: purge the form firewall records that keep IP addresses forever

### LINKED ISSUE #3682: form_firewall keeps IP addresses forever
The form firewall records the client IP address of every submission it watches, but nothing guarantees those records are ever removed. An IP address is personal data, so the table accumulates it with no retention limit.

### Where the rows come from

`FirewallForm::isFirewallOk()` inserts one row per (form, IP) pair and keeps it up to date:

- `core/lib/Thelia/Form/FirewallForm.php:69-72` — insert, storing `ip_address`
- `core/lib/Thelia/Form/FirewallForm.php:61` and `:64` — increment / touch

The column is `local/config/schema.xml:1687`, in the `form_firewall` table (`:1684-1696`), which carries a `timestampable` behavior.

### How long a row is actually useful

Only inside the waiting window. `isFirewallOk()` deletes everything older than `getConfigTime()` minutes before looking the caller up (`FirewallForm.php:46-52`), so a record past that point is never read again. The window defaults to 60 minutes (`FirewallForm.php:35`), or 10 for `BruteforceForm` (`BruteforceForm.php:26`). Anything older is an IP address kept for nothing.

### Why rows survive anyway

That cleanup at `:46-52` is opportunistic, and narrower than it looks:

- it is filtered by `filterByFormName($this::getName())`, so it only clears rows for the form currently being submitted — rows belonging to a form that is never submitted again (a form removed with a module, a rarely used one) are never reached;
- it runs only when `'prod' === $env && $this->isFirewallActive()` (`:42`), so switching `form_firewall_active` off freezes every existing row in place, permanently.

Nothing else deletes from the table: `FormFirewallQuery` declares no methods, and `MaintenancePurgeCommand` knows only three retention keys — carts without orders, anonymous carts, admin logs (`core/lib/Thelia/Command/MaintenancePurgeCommand.php:27-29`).

### Expected

`maintenance:purge` should drop `form_firewall` rows that are past the waiting window, the same way it already drops stale carts and admin logs. No schema change is needed — `updated_at` is already there.

A purge must not cut into the window the firewall still consults, otherwise it would hand a currently blocked IP address a fresh set of attempts.

anoziere:
Fixes #3682

`form_firewall` stores the client IP address of every submission the firewall watches (`local/config/schema.xml:1687`), and the only cleanup in the codebase is the opportunistic one inside `FirewallForm::isFirewallOk()` (`:46-52`). That one is filtered by form name and gated on `'prod' === $env && isFirewallActive()`, so rows belonging to a form that is never submitted again — or to any shop that turned the firewall off — stay in place indefinitely. `maintenance:purge` did not know about the table.

A record is only read back inside the waiting window: `isFirewallOk()` clears everything older before looking the caller up. Past that point it is an IP address kept for nothing, so it can be dropped.

### Changes

- `FormFirewallPurger`, following `AdminLogPurger` and `CartPurger`: delete rows whose `updated_at` is older than the retention.
- `maintenance:purge` runs it, reading `purification_form_firewall_days` (default 1), alongside the three retention keys it already handles.

The purge never deletes a record the firewall would still consult, whatever the configured retention: the threshold is clamped to the longest of `form_firewall_time_to_wait` and `form_firewall_bruteforce_time_to_wait`. Purging one early would hand a blocked IP address a fresh set of attempts.

No schema change — `updated_at` comes from the table's existing `timestampable` behavior.

### Tests

`FormFirewallPurgerTest` covers the three behaviors: expired rows go, recent rows stay, and rows inside the waiting window survive even a retention of 0 days. Each was checked red first — neutralizing the delete fails the first, dropping the clamp fails the third.

### DIFF
--- a/core/lib/Thelia/Command/MaintenancePurgeCommand.php
+++ b/core/lib/Thelia/Command/MaintenancePurgeCommand.php
@@ -20,6 +20,7 @@ use Thelia\Core\Event\Maintenance\MaintenancePurgeEvent;
 use Thelia\Core\Event\TheliaEvents;
 use Thelia\Domain\Admin\Service\AdminLogPurger;
 use Thelia\Domain\Cart\Service\CartPurger;
+use Thelia\Domain\Form\Service\FormFirewallPurger;
 use Thelia\Model\ConfigQuery;
 
 class MaintenancePurgeCommand extends ContainerAwareCommand
@@ -27,18 +28,22 @@ class MaintenancePurgeCommand extends ContainerAwareCommand
     private const DEFAULT_CART_NO_ORDER_DAYS_KEY = 'purification_cart_no_order_days';
     private const DEFAULT_CART_ANONYMOUS_DAYS_KEY = 'purification_cart_anonymous_days';
     private const DEFAULT_ADMIN_LOGS_DAYS_KEY = 'purification_admin_logs_days';
+    private const DEFAULT_FORM_FIREWALL_DAYS_KEY = 'purification_form_firewall_days';
 
     public function configure(): void
     {
         $this
             ->setName('maintenance:purge')
             ->setDescription(
-                'Purge old data from the database: carts without orders, anonymous carts, and admin logs.'
+                'Purge old data from the database: carts without orders, anonymous carts, admin logs, and form firewall records.'
             );
     }
 
-    public function __construct(protected AdminLogPurger $adminLogPurger, protected CartPurger $cartPurger)
-    {
+    public function __construct(
+        protected AdminLogPurger $adminLogPurger,
+        protected CartPurger $cartPurger,
+        protected FormFirewallPurger $formFirewallPurger,
+    ) {
         parent::__construct();
     }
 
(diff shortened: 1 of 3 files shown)


anoziere:
### Note on the red `PHPStan level 5 (BO Twig bundle)` check

This job was already failing on `main` before this branch existed, and it fails for a reason unrelated to these changes. Recording the evidence here.

The job analyses `templates/backOffice/default-twig/src`, a path this PR does not touch — the diff is limited to `core/lib/Thelia/Command/MaintenancePurgeCommand.php`, `core/lib/Thelia/Domain/Form/Service/FormFirewallPurger.php` and one test.

The same error appears on the branch base, on this PR, and on `main` after the merge:

```
Parameter #1 $array (list<int>) of array_values is already a list, call has no effect.
```

| Commit | Run | Result |
|---|---|---|
| `7e821eae4` — `main`, this branch's base | [94170015879](https://github.com/thelia/thelia/actions/runs/31613368851) | same error |
| this PR | [94377488475](https://github.com/thelia/thelia/actions/runs/31678234356) | same error |
| `d53ce82c8` — `main` after the merge | [94379989276](https://github.com/thelia/thelia/actions/runs/31679027603) | same error |

The job has been red on every `main` commit since 2026-08-12 (`5fca22238`, `01f8d0361`, `7e821eae4`, `da43bf906`, `64f4477ff`).

The gating jobs were green: `CI (coding standard, static analysis, tests)` passed on both 8.3 and 8.4. Scrutinizer was left running and did not gate the merge.

Worth fixing on its own: while this job stays red, it cannot signal a real regression in the back-office bundle.

Issue #3520, opened 2026-08-11
Rethink the virtual product / file association (one file per PSE, stored in meta_data)


anoziere:
Split from https://github.com/thelia/thelia/issues/2002 (its two remaining bullets). Current design, unchanged since Thelia 2: the association between a product sale element and its downloadable file is a single `meta_data` row (`virtual` key, `pse` element), so one file per combination, no referential integrity, and no room for multiple files per PSE.

Two questions to arbitrate for Thelia 3 (schema changes are still cheap in beta):
1. Should the association move to a real table (FK to `product_sale_elements` and `product_document`) instead of `meta_data`?
2. Should it support several files per PSE?

Note the per-combination selection already exists (the original report's premise that only one file per product is supported was wrong); the actual gaps are storage robustness and multi-file support.

anoziere:
Decision for 3.0: the association stays in `meta_data` (`virtual` key, `pse` element). Two reasons. The beta is close to a stable release and a storage change means a migration plus a data backfill on every install that already sells virtual products. And the key is read directly by modules outside the core (any virtual delivery module that looks up the file for a combination), so moving it silently breaks them without a deprecation window.

What is being fixed now is the back office, which was reading and writing the wrong place: see #3516. After that, the Twig back office and the legacy Smarty one agree on the same storage, and the documented behaviour matches the code.

The two questions in this ticket stay open for after 3.0, and they are worth separating:

1. A real table with foreign keys to `product_sale_elements` and `product_document`. This is the one that buys referential integrity: today a deleted document leaves a dangling id in `meta_data` and nothing notices.
2. Several files per combination. This one needs a product answer first (does the customer get all files, or picks one?), and it changes the order side too, since `order_product.virtual_document` holds a single file name.

Keeping this ticket as the design entry point for both, with the `needs-decision` label dropped since the 3.0 part is settled.

anoziere:
## Who reads and writes the key today, and what a move would cost

The 3.0 arbitration above stands. What was missing to decide the rest is the list of everything that
touches the key, so here it is, followed by a concrete migration plan for after 3.0.

### Storage

`meta_data` (`local/config/schema.xml:1916-1929`), one row per sale element:

| column | value |
|---|---|
| `meta_key` | `virtual` |
| `element_key` | `pse` (`Thelia\Model\MetaData::PSE_KEY`, `core/lib/Thelia/Model/MetaData.php:23`) |
| `element_id` | `product_sale_elements.id` |
[...]

PR #3687, opened 2026-08-13
feat: expose the virtual documents of a product sale element

### LINKED ISSUE #3686: Expose the virtual documents of a product sale element
The document a customer downloads after buying a virtual product is associated to a
`ProductSaleElements` through a `meta_data` row (`meta_key = virtual`, `element_key = pse`,
`value` = a `product_document.id`). There is no accessor for it: every reader has to know
that storage and query it directly, which is what `VirtualProductDelivery` does today.

#3520 plans to move that association to a dedicated table once 3.0 is out. A module reading
`meta_data` will break on that day; a module reading an accessor will not.

Add the accessor now, so the storage change becomes a core-only concern:

- `ProductSaleElements::getVirtualDocument(): ?ProductDocument`
- `ProductSaleElements::getVirtualDocuments(): ObjectCollection` — plural, because #3520 leaves
  the door open to several documents per sale element

The accessor also drops an id whose document has been deleted. Nothing cleans `meta_data` when
a `ProductDocument` goes away, so a dangling id currently surfaces downstream as a virtual
product with no file behind it.

No schema change, no behaviour change for existing readers: `meta_data` keeps being written and
read exactly as before. This is step 1 of the deprecation window described in #3520, which stays
open as the design entry for the storage migration.

anoziere:
The document behind a virtual product is associated to a `ProductSaleElements` through a
`meta_data` row, and every reader has to know that storage to reach it. This adds the accessor
that hides it:

- `getVirtualDocument(): ?ProductDocument` — the single document a sale element holds today
- `getVirtualDocuments(): ObjectCollection` — plural, because #3520 leaves the door open to
  several documents per sale element

Both drop an id whose document has been deleted: nothing cleans `meta_data` when a
`ProductDocument` goes away, so a dangling id otherwise travels downstream as a virtual product
with no file behind it.

No schema change and no behaviour change for existing readers — `meta_data` keeps being written
and read as before. This is step 1 of the deprecation window described in #3520, which stays open
as the design entry for the storage migration.

Fixes #3686

### DIFF
--- a/core/lib/Thelia/Model/ProductSaleElements.php
+++ b/core/lib/Thelia/Model/ProductSaleElements.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 
 namespace Thelia\Model;
 
+use Propel\Runtime\Collection\ObjectCollection;
 use Propel\Runtime\Connection\ConnectionInterface;
 use Propel\Runtime\Exception\PropelException;
 use Thelia\Domain\Taxation\TaxEngine\TaxCalculatorResolverTrait;
@@ -25,6 +26,14 @@ class ProductSaleElements extends BaseProductSaleElements
 {
     use PositionManagementTrait;
     use TaxCalculatorResolverTrait;
+
+    /**
+     * The meta data key holding the virtual document association.
+     *
+     * Read it through getVirtualDocuments() rather than directly: the storage is
+     * meant to move to a dedicated table, the accessor is not.
+     */
+    public const VIRTUAL_DOCUMENT_META_KEY = 'virtual';
 
     /**
      * @throws PropelException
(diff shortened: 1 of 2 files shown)


anoziere:
The `PHPStan level 5 (BO Twig bundle)` job is red on this pull request for a reason it does not
introduce. It fails identically on `main` — `Parameter #1 $array (list<int>) of array_values is
already a list, call has no effect.` — at 7e821ea (the base of this branch, before this change)
and on every `main` push since #3671:

| run | head | conclusion |
|---|---|---|
| https://github.com/thelia/thelia/actions/runs/31613368851 | 7e821ea (base) | failure |
| https://github.com/thelia/thelia/actions/runs/31612904566 | 5fca222 (#3671) | failure |

This branch touches no file under `templates/backOffice/default-twig/`. The `CI (coding standard,
static analysis, tests)` job, which runs cs-diff, PHPStan on `core` and the five PHPUnit suites,
passes on 8.3 and 8.4.

PR #3688, opened 2026-08-13
fix: stop Customer::getLocale() from fatalling when the account has no language

### LINKED ISSUE #3669: Demo customers have a null lang_id, which makes Customer::getLocale() fatal on a stock install
Reported twice today by two independent sessions, and the second one established why it is reachable on a stock install rather than only in tests.

`core/lib/Thelia/Model/Customer.php:353-356` — `getLocale()` calls `$this->getLangModel()->getLocale()` with no guard, while `customer.lang_id` is nullable. Any customer created without a language produces `Call to a member function getLocale() on null`.

That was first hit through `FixtureFactory::customer()`, which does not set `lang_id`, and could have been dismissed as a test-fixture problem. It is not: **the demo customers ship with `customer.lang_id = NULL`**, so the defect is reachable on any install with demo data, through any code path that asks a customer for its locale — which the personal-data exporter does, and which the mailer does when it picks the language of a customer notification.

The fix is a guard returning the default language locale when the customer carries none, and it is worth checking whether other nullable foreign keys are dereferenced the same way on that model.

anoziere:
Closes #3669.

`customer.lang_id` is nullable by design, and the `lang` foreign key is declared `onDelete="SET NULL"`, so deleting a language in the back-office empties the column for every customer that used it. `Customer::getLocale()` dereferenced that relation without a guard and fatalled with `Call to a member function getLocale() on null`, while the interface it implements (`Thelia\Core\Security\User\UserInterface::getLocale(): string`) promises a string.

No core caller hits it today — they all work around it, one of them with a comment describing this very bug — but any module or theme coding against the interface does.

**Return the default locale when the customer has no language**

`getLocale()` now goes through `getCustomerLang()`, which already implements that fallback, so the rule lives in one place. `getCustomerLang()` itself switches to `Lang::getDefaultLanguage()`: its `Lang` return type was a lie, since `findOne()` returns `null` on a shop with no default language, and the helper raises an explicit message instead of a `TypeError`. The obsolete comment in `CustomerPersonalDataExporter` is reworded — the export deliberately reports the language the account carries, not the fallback.

**Give demo customers the default language**

`CustomersImporter` never set `lang_id`, so `--with-demo` produced ten accounts without a language. It now resolves the default language once and assigns it.

No schema change and no migration for existing databases: `NULL` keeps its meaning of "no preference, follow the shop default", and with the guard in place it is safe again.

Verified red before the fix and green after: the new `CustomerLocaleTest` reproduced the exact fatal, and the demo-import assertion reported the ten customers.

Baselines: 858 tests green across the five suites, `cs-diff` 0/1761, PHPStan 0 errors.

### DIFF
--- a/core/lib/Thelia/Command/Import/Importer/CustomersImporter.php
+++ b/core/lib/Thelia/Command/Import/Importer/CustomersImporter.php
@@ -18,6 +18,7 @@ use Thelia\Command\Import\AbstractDemoImporter;
 use Thelia\Command\Import\DemoImportContext;
 use Thelia\Model\Address;
 use Thelia\Model\Customer;
+use Thelia\Model\LangQuery;
 
 /**
  * Creates demo customers and their addresses directly through Propel models.
@@ -42,6 +43,11 @@ final class CustomersImporter extends AbstractDemoImporter
 
     public function import(DemoImportContext $context): void
     {
+        $defaultLangId = LangQuery::create()
+            ->filterByByDefault(1)
+            ->findOne($context->connection)
+            ?->getId();
+
         $isFirst = true;
         foreach ($this->readCsv($context->dataDir.'customers.csv') as $data) {
             $titleId = (int) $data[0];
@@ -52,6 +58,7 @@ final class CustomersImporter extends AbstractDemoImporter
             $customer->setLastname($data[2]);
             $customer->setEmail($data[3]);
             $customer->setPassword(self::DEMO_PASSWORD);
+            $customer->setLangId($defaultLangId);
             $customer->save($context->connection);
 
             (new Address())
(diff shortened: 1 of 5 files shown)


anoziere:
The "PHPStan level 5 (BO Twig bundle)" check is red here for a reason unrelated to this branch, and it is red on `main` as well.

The job analyses `templates/backOffice/default-twig/src`, which it re-clones from its own repository, and fails on:

```
Parameter #1 $array (list<int>) of array_values is already a list, call has no effect.
```

Same message and same job on the base branch:

- `main` @ da43bf90 — https://github.com/thelia/thelia/actions/runs/31678373609
- `main` @ 7e821eae, the commit this branch starts from — https://github.com/thelia/thelia/actions/runs/31613368851

The job has been failing on `main` since 5fca2223 (2026-08-12). This pull request touches no file it analyses.

The two CI jobs that cover this change — coding standard, static analysis and the full test suite on PHP 8.3 and 8.4 — are green.

PR #3689, opened 2026-08-13
fix: export the shipping tax rule title as text, and the delivery country in its own column

### LINKED ISSUE #3667: The accounting order export rounds a VARCHAR title to two decimals
Found while mapping the postage readers for #1664.

`core/lib/Thelia/Domain/DataTransfer/Export/Type/OrderExport.php:96` selects `ROUND(`order`.postage_tax_rule_title, 2)`. That column is the **title** of the tax rule, a VARCHAR — rounding it to two decimals makes MySQL coerce a string to a number, so the accounting export ships `0.00` where the rule name should be, for every order.

The same shape exists on `2.6` in `core/lib/Thelia/ImportExport/Export/Type/OrderExport.php`.

It is in the export accountants read, so the wrong value has been leaving the shop for a while without anyone reading the SQL. Worth checking the neighbouring columns of that query at the same time, since a `ROUND()` applied to the wrong column suggests a copy of the line above.

anoziere:
The accounting order export carries two defects that both come from the same
copy-and-paste, and both are fixed here. Same change as #3680, ported to `main`
(the two files have diverged, so no cherry-pick).

**`ROUND()` on a VARCHAR** — `core/lib/Thelia/Domain/DataTransfer/Export/Type/OrderExport.php:96`
selected `ROUND(\`order\`.postage_tax_rule_title, 2)`. That column is a
`VARCHAR(255)` (`local/config/schema.xml:624`), so MySQL coerced the label to a
number and every exported order carried `0` instead of the tax rule name —
`SELECT ROUND('TVA 20%', 2)` returns `0.00`. The line dates back to 2020
(`03c0bd99c8`); the two lines around it round `discount` and `postage`, which
are real `DECIMAL(16,6)` columns.

**Alias collision on the delivery country** — `:57` and `:68` both mapped to the
output alias `invoice_country`. `applyOrderAndAliases()` writes the aliases in
order (`JsonFileAbstractExport.php:96-106`), so the invoice country overwrote the
delivery country and the latter never reached the file. Every other entry in the
surrounding block is already prefixed `delivery_`.

Note for integrators: the output gains a `delivery_country` column, in its
position in the column order. `invoice_country` keeps exactly the value it has
today, so no consumer loses data — only readers that index columns by position
need to be aware.

Two cases were added to `tests/Integration/Domain/DataTransfer/OrderExportTest.php`,
along with an `exportedRow()` helper next to the existing `exportedRefs()`. Both
were checked red against the current code (`0` instead of `TVA 20%`, and no
`delivery_country` key) and green with the fix.

Fixes #3667

### DIFF
--- a/core/lib/Thelia/Domain/DataTransfer/Export/Type/OrderExport.php
+++ b/core/lib/Thelia/Domain/DataTransfer/Export/Type/OrderExport.php
@@ -54,7 +54,7 @@ class OrderExport extends JsonFileAbstractExport
         'delivery_address_address3' => 'delivery_address_3',
         'delivery_address_zipcode' => 'delivery_zip_code',
         'delivery_address_city' => 'delivery_city',
-        'delivery_country_i18n_title' => 'invoice_country',
+        'delivery_country_i18n_title' => 'delivery_country',
         'delivery_address_phone' => 'delivery_phone',
         'invoice_address_customer_title_long' => 'invoice_title',
         'invoice_address_company' => 'invoice_company',
@@ -93,7 +93,7 @@ class OrderExport extends JsonFileAbstractExport
                     order_coupon.code as order_coupon_code,
                     ROUND(`order`.postage, 2) as order_postage,
                     `order`.postage_tax as "order_postage_tax",
-                    ROUND(`order`.postage_tax_rule_title,2) as "order_postage_tax_rule_title",
+                    `order`.postage_tax_rule_title as "order_postage_tax_rule_title",
                     SUM(ROUND(order_product.quantity * IF(order_product.was_in_promo = 1, order_product.promo_price, order_product.price), 2) ) as order_total_price,
                     SUM(
                         ROUND(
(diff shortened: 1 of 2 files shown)


anoziere:
The `PHPStan level 5 (BO Twig bundle)` check is red here for a reason unrelated
to this PR, which touches nothing under `templates/backOffice/default-twig/src`.

The failure is `templates/backOffice/default-twig/src/Repository/ModuleRepository.php:82`
— *Parameter #1 $array (list<int>) of array_values is already a list, call has no
effect*. The same job fails with the same single error on the base branch itself,
on the push runs that built `main` right before and right after this PR was
opened:

- `main` @ `da43bf906c` — https://github.com/thelia/thelia/actions/runs/31678373609
- `main` @ `d96af578be` — https://github.com/thelia/thelia/actions/runs/31679217527
- this PR — https://github.com/thelia/thelia/actions/runs/31678532267

It is a pre-existing break on `main` and needs its own fix.

Issue #3168, opened 2024-02-09
Adding iso 3166-2 support for states


bvey:
### Description

Thelia state datatable contains only two-character iso codes (ISO 3166 standard).

Adding the five-character iso code support (ISO 3166-2).

For the US states : https://en.wikipedia.org/wiki/ISO_3166-2:US

anoziere:
Triage notes. The full `US-CA` form is already derivable today (`country.isoalpha2` + `state.isocode`), so the actionable gap is convenience, plus the much larger question of worldwide coverage (the seed only ships subdivisions for 7 countries).

Three options weighed:

**A. Computed accessor (shipping now).** `State::getIsoCode3166_2()` returning `{country alpha-2}-{state isocode}`, exposed as a new `ISOCODE_3166_2` output in the state loop and as a read-only field on the API resource. No schema change, no migration.

**B. Repurpose `state.isocode` to store the full code.** Rejected: breaking for every consumer of the bare suffix (the `/admin/states/iso/{countryIso3}/{stateIso}` API route, the address validation traits, 312 seeded rows) for a purely cosmetic gain.

**C. Full worldwide ISO 3166-2 seed (~5,000 rows).** That part is real but belongs in an opt-in geo-data module rather than a 150x growth of the core seed — a module can also track ISO 3166-2 changes independently of Thelia releases. Kept open here as the roadmap discussion.

A is on its way for both branches.

anoziere:
Option A is merged on both lines: #3530 (`2.6`) and #3531 (`main`). `State::getIsoCode3166_2()` returns the full `US-CA` form (null-safe), exposed as `ISOCODE_3166_2` in the state loop and as a read-only `isoCode3166_2` field on the API resource. This issue stays open for the remaining question: worldwide ISO 3166-2 seed coverage (~5,000 rows), which we recommend shipping as a dedicated opt-in geo-data module rather than in the core seed.

anoziere:
## What the seed actually ships today

The remaining half of this issue is a data question, and the starting figure was never written down anywhere. Here it is, counted from `setup/insert.sql` on both lines and compared against the ISO 3166-2 register (`iso-codes` 4.15.0, 2024-01, 5046 entries).

| | `main` (Thelia 3) | `2.6` |
|---|---|---|
| countries seeded | 249 | 249 |
| countries carrying subdivisions | **8** | **7** |
| subdivision rows | **413** | **312** |
[...]

PR #3690, opened 2026-08-13
fix: finish the ISO 3166-2 clean-up of the state seed

anoziere:
Two data defects left by the ISO 3166-2 work on #3168.

**Ciudad de México kept its pre-2016 name.** The entity was renamed at the same time it was recoded `CMX`. The Spanish title already said Ciudad de México, the English, French and Russian ones still read Distrito Federal, so the same state answered to two names depending on the language of the shop. The source string in `insert.sql.tpl` and the `setup/I18n` key are renamed as well, otherwise the next regeneration of `insert.sql` would bring the old name back.

**Four Italian provinces dissolved in 2016.** Carbonia-Iglesias, Medio Campidano, Ogliastra and Olbia-Tempio were merged into Sud Sardegna and Sassari, and ISO 3166-2:IT withdrew `IT-CI`, `IT-VS`, `IT-OG` and `IT-OT`. They are seeded `visible = 0` rather than deleted, so an address already pointing at one of them stays readable, and `IT-SU` (Sud Sardegna) is added so Sardinia stays selectable.

Both changes come with a guarded statement in `3.0.0-beta2.sql` for installed shops: matched on the outdated value only, so a shop that fixed it by hand is left alone and a second run is a no-op.

Worth knowing before merging: `visible = 0` removes the four rows from the `state` loop, which defaults to `visible="yes"`, but not from the address form. `AddressCreateForm::getStatesChoices()` queries `StateQuery` without filtering on `visible`, so a state hidden in the back office is still offered to the customer today. That is a separate defect and this pull request does not touch it.

Left as they are, on purpose:

- `IT-AO` — Aosta is a province code the standard does define, it was wrongly listed as invalid.
- `FR-75` — ISO codes Paris `FR-75C`; the French departments seeded by #3536 deliberately use the INSEE numbering, and changing this one would make France the only country mixing two schemes.

Refs #3168


anoziere:
`PHPStan level 5 (BO Twig bundle)` is red on this branch and red on `main` without it: run [31679027603](https://github.com/thelia/thelia/actions/runs/31679027603) on `main` (d53ce82) fails with the same single error, `Parameter #1 $array (list<int>) of array_values is already a list, call has no effect.` This pull request only changes SQL seed files, translation arrays and one test, and touches nothing under `templates/backOffice/default-twig/src`. The two `CI (coding standard, static analysis, tests)` jobs, 8.3 and 8.4, pass.

PR #3691, opened 2026-08-13
fix: carry the virtual document of each sale element when cloning a product

### LINKED ISSUE #3662: Cloning a virtual product marks the clone virtual without a file
Found while mapping the virtual product storage for #3520.

`Action/Product::updateClone()` (`core/lib/Thelia/Action/Product.php:196`) never calls `setVirtualDocumentId()`. Cloning a virtual product therefore produces a product flagged as virtual whose sale elements point at no file: the customer buys it, and the delivery module finds nothing to send.

The fix is not a one-liner, which is why it was left out: the clone gets its own documents through `FILE_CLONE`, so the source document id has to be remapped to the newly created one, and the `meta_data` row written with the new id.

Worth deciding at the same time what should happen when the source product is virtual but has no file attached — mark the clone virtual anyway, or not.

anoziere:
Cloning a virtual product produced a clone still flagged `virtual = 1`, but with no `meta_data` row telling which document the customer downloads. The order then went through — `VirtualProductHandler::resolve()` (`core/lib/Thelia/Domain/Order/Service/VirtualProductHandler.php:35-42`) dispatches on the flag alone — while `VirtualProductDelivery` found no association, so `order_product.virtual_document` stayed empty and the customer paid for nothing.

Fixes #3662

## Why it could not be fixed where the flag is copied

`updateClone()` (`core/lib/Thelia/Action/Product.php:196-246`) runs before `FILE_CLONE` (`:161`), so the clone has no documents yet and no id to point at. The only place holding both the source and cloned sale element ids, with the cloned documents already created, is `PSE_CLONE`.

The missing link was the source document id to cloned document id correspondence: `File::cloneFile()` had it in memory but never exposed it.

## Changes

- `core/lib/Thelia/Core/Event/Product/ProductCloneEvent.php` — carries the map, `addClonedDocumentId()` / `getClonedDocumentId()`. Additive, no signature changed.
- `core/lib/Thelia/Action/File.php:124-127` — fills it right after `DOCUMENT_SAVE`, where the copy has its id.
- `core/lib/Thelia/Action/ProductSaleElement.php:394` — `clonePSE()` copies the `virtual`/`pse` `meta_data` row of every sale element, remapped onto the clone's own document.

Two deliberate rules:

- The association is copied whenever the source has one, whether or not the product is currently flagged virtual, so a temporarily unchecked flag does not silently destroy data on clone.
- When the source document was not copied (its file is missing from the disk), the clone gets no association and a `Tlog` warning is logged, rather than a pointer to a file that will be deleted with the original product. The `virtual` flag is kept: it carries a merchant intent, and the clone is created invisible anyway.

Position matching, as `clonePSEAssociatedFiles()` (`:480-484`) does for regular file associations, was rejected: `position` is not unique per product, and a source document skipped for a missing file shifts the match onto the wrong document.

No schema change.

## Tests

`tests/Integration/Action/ProductSaleElementActionTest.php`

- `testCloneCarriesTheVirtualDocumentOfEachSaleElement()` — a virtual product with two combinations, a real file on disk, one association per sale element; asserts each cloned sale element points at a document that exists, belongs to the clone, and is not the source one. Red without the fix.
- `testCloneLeavesNoVirtualAssociationWhenTheSourceFileIsMissing()` — locks the behaviour above; red if the fix falls back to the source document id.

Both clean up the media directory in `tearDown()`, which the database rollback does not cover.

The same defect exists on 2.6; this PR only covers the Thelia 3 line.

### DIFF
--- a/core/lib/Thelia/Action/File.php
+++ b/core/lib/Thelia/Action/File.php
@@ -121,6 +121,10 @@ class File extends BaseAction implements EventSubscriberInterface
                         case 'documents':
                             $dispatcher->dispatch($clonedProductCreateFileEvent, TheliaEvents::DOCUMENT_SAVE);
 
+                            // Keep track of which copy replaces which source document, so the
+                            // virtual document of each sale element can be remapped later on.
+                            $event->addClonedDocumentId((int) $originalProductFile->getId(), (int) $clonedProductFile->getId());
+
                             // Get original product document I18n
                             $originalProductFileI18ns = ProductDocumentI18nQuery::create()
                                 ->findById($originalProductFile->getId());
--- a/core/lib/Thelia/Action/ProductSaleElement.php
+++ b/core/lib/Thelia/Action/ProductSaleElement.php
@@ -20,6 +20,7 @@ use Propel\Runtime\Exception\PropelException;
 use Propel\Runtime\Propel;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Thelia\Core\Event\MetaData\MetaDataCreateOrUpdateEvent;
 use Thelia\Core\Event\Product\ProductCloneEvent;
 use Thelia\Core\Event\Product\ProductCombinationGenerationEvent;
 use Thelia\Core\Event\ProductSaleElement\ProductSaleElementCreateEvent;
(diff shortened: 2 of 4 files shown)


anoziere:
`PHPStan level 5 (BO Twig bundle)` is red here for a reason unrelated to this branch, which touches no file under `templates/backOffice/default-twig/src` (the only path that job analyses).

Same failure, same message — `Parameter #1 $array (list<int>) of array_values is already a list, call has no effect.` — on `main` itself:

- base commit of this PR, `7e821ea`: [run 31613368851](https://github.com/thelia/thelia/actions/runs/31613368851)
- current `main`, `d53ce82`: [run 31679027603](https://github.com/thelia/thelia/actions/runs/31679027603)

The job turned red on `main` at `5fca222` (#3671, 12/08); the run just before it, `e5ba7a0`, was green.

Issue #3589, opened 2026-08-12
GDPR: retention, entry points and the marker column the anonymisation path still needs


anoziere:
Follow-up to #2460, once #3588 lands the core anonymisation and personal data export path.

#3588 gives a module the two operations it needs: `TheliaEvents::CUSTOMER_ANONYMIZE`, `TheliaEvents::CUSTOMER_PERSONAL_DATA_EXPORT`, and `CustomerPersonalDataProviderInterface` so a module can declare the personal data it holds. Four things were deliberately left out of that pull request, each for a reason.

### 1. A marker recording that an account was anonymised — needs a schema change

There is no way to tell an anonymised account from an ordinary one, other than matching the placeholder email, which is a heuristic and not a fact. A retention job cannot skip the accounts it has already processed, the back-office cannot display the state, and a second pass cannot be made idempotent by anything other than convention.

The fix is one nullable `DATETIME` column, `customer.anonymized_at`. It was not added in #3588 because a change to `local/config/schema.xml` fails the CI of `main` until `thelia/config` has been dispatched and released, and these changes are batched.

### 2. Retention

`MaintenancePurgeCommand` already purges carts (`purification_cart_no_order_days`, `purification_cart_anonymous_days`) and admin logs (`purification_admin_logs_days`) — see `core/lib/Thelia/Command/MaintenancePurgeCommand.php:27-29`. Customers and orders have no equivalent. Once the column of point 1 exists, a job can anonymise accounts inactive for longer than a configured retention period, driven by the same command.

### 3. Entry points

Neither the back-office nor the front-office can trigger the two operations. `CUSTOMER_DELETEACCOUNT` itself has no dispatcher in this repository, and `core/lib/Thelia/Action/Customer.php:202-211` still refuses deletion when orders exist, which stays the correct behaviour.

- back-office: an "anonymise" action next to the delete action of a customer, and a download of the archive — `thelia-templates/default-twig`
- front-office: self-service "download my data" and "delete my account" on the account page — `thelia-templates/flexy`

### 4. Residual personal data in `admin_log`

`AdminLog::append()` stores the serialised request, POST content included (`core/lib/Thelia/Model/AdminLog.php:28-50`). A back-office edit of a customer therefore leaves their name, email and address in `admin_log.request`. Anonymisation does not touch it: it is the administrator audit trail, and deleting rows from it on behalf of a customer would destroy evidence unrelated to that customer. The retention period above is the right lever, but the trade-off deserves an explicit decision.

[...]

PR #3692, opened 2026-08-13
feat: mark an anonymized customer account, and let an order outlive its module

anoziere:
Two schema changes that both need a `thelia/config` release, grouped so the shop
pays that cycle once, plus the GDPR code that the first of them unblocks.

Closes part of #3589. Prepares #1145.

## Schema

**`customer.anonymized_at`** (+ `idx_customer_anonymized_at`) — #3589 asked for a
marker. Without one, nothing separates an anonymous account from an account that
simply has no name yet: the back office cannot show the state, the operation can
be replayed silently, and a retention job has no way to skip the accounts it
already handled. The index is what that job filters on.

**`order.payment_module_id` and `order.delivery_module_id` become nullable** —
today both are `NOT NULL` behind a `RESTRICT` foreign key, so a module used by a
single order can never be deleted. Dropping `NOT NULL` is the structural half of
letting an order outlive its module. `cart` already does exactly this on the same
two columns. No existing row changes; every order keeps the module ids it has.

`customer` and `order` are both versionable, so `customer_version` and
`order_version` are altered the same way.

## Migration

Appended to `setup/update/sql/3.0.0-beta2.sql`, the script of the release being
prepared — the last schema changes landed the same way. Every statement is
guarded on `information_schema`, so a shop installed from a `main` snapshot is
left alone and a second run is a no-op.

Checked against a real MariaDB: `setup/thelia.sql` from `main` loaded into one
database and migrated, `setup/thelia.sql` from this branch loaded into another,
then the full `information_schema.COLUMNS` of both compared — identical, and the
script replayed cleanly.

## Code (#3589)

- `CustomerAnonymizer` stamps the account. It records the *first* erasure:
  replaying the operation must not push the date forward, or the retention job
  loses track of when the data actually went away.
- `CustomerPurger` applies a retention period. An account is measured against
  two distinct periods, because an account backed by orders is also backed by an
  accounting retention obligation: an account that never ordered ages from the
  day it was opened, an account that ordered ages from its most recent order.
  Anonymization goes through `CUSTOMER_ANONYMIZE`, so the modules listening to
  it erase their own data on a retention run exactly as they do when an
  administrator triggers the operation by hand.
- Both periods default to **0, meaning off**. Erasing an identity cannot be
  undone and the job runs unattended; only the shop knows how long it is allowed
  to keep the data.
- `maintenance:purge --dry-run` reports what the run would remove without
  touching anything. The `form_firewall` purge of #3684 landed first, so this
  branch is rebased on it and wires `FormFirewallPurger` into `--dry-run` too,
  keeping the flag honest for every purger the command runs.

## CI is expected to fail here, and to go green on its own

On `main` the workflow extracts `thelia/config` into `local/config/`, which
overwrites `schema.xml` **before** the Propel models are generated. The new
column therefore does not exist for the analysis step, and PHPStan reports
undefined methods on `getAnonymizedAt` / `setAnonymizedAt` /
`filterByAnonymizedAt`. The integration tests covering the marker fail for the
same reason.

This clears once `thelia/config` carries the schema and is tagged. Re-running the
workflow after that release is all it takes.

## Not in this PR

- The rest of #1145 — freezing the module title and code on the order, the API
  resource, the `INNER JOIN`s that assume a module is always there, and the
  deletion path itself. That code needs the nullable columns to exist first, so
  it follows the config release.
- The `admin_log` residue (#3589 point 4), which is still an open arbitration.
- No Propel stub or `Base/` file is committed: generated models are not tracked
  in this repository, and the precedent for adding a column (#3533) touched only
  `schema.xml`, `setup/thelia.sql` and the business code.


anoziere:
### CI result, checked failure by failure

`cs-diff`, PHPStan (`[OK] No errors`) and the unit suite (279 tests) pass on both
PHP 8.3 and 8.4. `PHPStan level 5 (BO Twig bundle)` passes.

The integration suite reports 12 errors and 1 failure, identical on 8.3 and 8.4,
and every one of them is the same single cause — the column missing from the
Propel models CI generates:

- 6 × `BadMethodCallException: Call to undefined method: getAnonymizedAt` /
  `setAnonymizedAt`, raised from
  `var/propel/test/model/Thelia/Model/Base/Customer.php`
- 6 × `UnknownColumnException: Unknown column AnonymizedAt in model Thelia\Model\Customer`
- the single "failure" is the same exception surfacing in
  `testAnonymizeStopsWhenAProviderFails`, where a `RuntimeException` was expected

No other exception or error appears anywhere in either job.

The cause is the one this PR was opened with: `composer install`/`update` installs
`thelia/config`, whose `schema.xml` replaces `local/config/schema.xml` before
`bin/test-prepare` builds the models, so the models are built from a schema
without `anonymized_at`. Nothing in the diff can fix that from this repository.

Merging on that basis. Once `thelia/config` carries the column and is tagged,
re-running this workflow on `main` is all it takes.

One caveat for that re-run: `composer ci` stops at the first failing suite, so the
api, flexy and backoffice suites were never reached here. They should be green —
`anonymized_at` is not exposed on any API resource and no template reads it — but
the post-release run is what actually confirms it.

PR #3698, opened 2026-08-13
fix: give cart.postage and order.postage the same meaning

### LINKED ISSUE #3668: cart.postage and order.postage carry the same name and the opposite meaning
Found while mapping the postage readers for #1664, and it is the kind of trap that produces a wrong total years later.

`Action/Cart.php:146` stores `amount - amountTax` in `cart.postage` — **tax exclusive**. `Checkout/Service/CheckoutPaymentService.php:51` adds the tax back when building the order, so `order.postage` is **tax inclusive**. Which is why `Order::getUntaxedPostage()` (`Model/Order.php:364`) has to subtract it again.

Two columns, the same name, opposite conventions, and nothing in either name or docblock says so. Every new reader has a one-in-two chance of getting it wrong, and the error is invisible until someone compares a cart total with an order total on a shop where postage is taxed.

This matters now because the ventilation designed in #1664 has to know which side of the convention it computes on. At minimum the two should be documented at their definition; better, one of them should be renamed so the convention is legible where it is read.

anoziere:
`cart.postage` and `order.postage` carry the same name, the same Propel type and the opposite meaning: the cart column was stored tax **excluded**, the order column tax **included**. Nothing said so — no `description` in the schema, no docblock at either definition — and both models shipped a mirror helper to undo the other one's convention (`Cart::getTaxedPostage()` adds the tax, `Order::getUntaxedPostage()` removes it).

A single postage travelled through two conversions:

```
DeliveryModule::getPostage()  ->  OrderPostage(amount=12.00, amountTax=2.00)   tax included
Action\Cart::calculatePostage ->  cart.postage = 10.00, cart.postage_tax = 2.00  tax excluded
CheckoutPaymentService::pay   ->  order.postage = 12.00, order.postage_tax = 2.00  tax included
```

The core and the shipped themes were correct, because they all went through the helpers. Readers that did not — and there are several, in modules — got one chance in two.

This converges the cart on the convention every other part of the system already applies: the `OrderPostage` value object, the `order` and `delivery` loops, the `{postage}` Smarty function, the PDF invoice, the order emails, the back office, and ten years of `order` rows. Going the other way was never an option: it would break every delivery and payment module, every invoice and all historical data.

## What changed

`cart.postage` now holds the amount the delivery module quoted, tax included, exactly like `order.postage`. `cart.postage_tax` is unchanged.

- `Action\Cart::calculatePostage()` writes `OrderPostage::getAmount()` instead of `getAmount() - getAmountTax()`.
- `Cart::getTaxedPostage()` returns the column as-is; a new `Cart::getUntaxedPostage()` mirrors `Order::getUntaxedPostage()`.
- `Cart::getTotalAmount(withPostage: true)` adds the untaxed postage, so the tax-excluded total stays tax-excluded.
- `CheckoutPaymentService::pay()` copies the cart postage to the order instead of re-adding the tax.
- `PostageEstimator::estimatePostageForCountry()` returns the tax-included amount, so `DeliverySetupService` keeps writing a consistent cart and `ShippingFacade::estimateCartPostage()` keeps returning the same `PostageEstimateView(amountHt, tax, totalTtc)` triple.
- `OrderQuery::getSaleStats()` no longer adds a tax-included postage to a tax-excluded turnover.

## For integrators

**Read `cart.postage` directly and you now get a tax-included amount.** Code that used to add `cart.postage_tax` on top of it — because that is what the cart demanded — now counts the shipping VAT twice.

| Expression | Before | After |
|---|---|---|
| `$cart->getPostage()` | tax excluded | tax included |
| `$cart->getPostage() + $cart->getPostageTax()` | the amount to charge | shipping VAT counted twice: call `getPostage()` alone |
| `$cart->getTaxedPostage()` | the amount to charge | the amount to charge, unchanged |
| `$cart->getUntaxedPostage()` | did not exist | the tax-excluded amount |

`Cart::getTaxedPostage()` keeps working and stays the safest thing to call: it returned the customer-facing amount before this change and it still does.

Data access and API:

- `attributeCart('postage')` becomes tax included; `attributeCart('taxed_postage')` is unchanged; `attributeCart('untaxed_postage')` and `attributeCart('postage_tax')` are new. The cart attribute set is now the exact mirror of the order one.
- `GET /api/front/carts/{id}` returns `delivery` tax included. It is now consistent with `total`, which is tax included too — adding the two used to under-charge the shipping VAT.

No migration is needed: `cart.postage` is recomputed on every cart mutation (`PostageHandler::handlePostageOnCart()`), so a cart that is touched is a cart that is repriced. A shop that wants existing carts realigned right away can run `UPDATE cart SET postage = postage + postage_tax WHERE postage_tax > 0;` once.

## Tests

`tests/Integration/Domain/Checkout/CartOrderPostageSemanticsTest.php` walks a taxed postage from the delivery quote to the placed order through the real `CART_SET_POSTAGE` listener and a real `CheckoutPaymentService::pay()`, and asserts the amount never changes on the way. Without this fix it reports `cart.postage = 10.00` against `order.postage = 12.00`.

## Out of scope

- The `description` attributes proposed on the six schema columns are left for a batched `local/config/schema.xml` change, since each one costs a `thelia/config` release cycle.
- Seven modules read the columns on the wrong basis today (PayPal, Shopimind, AdminOrderCreation, Statistic, TheliaGiftCard, GoogleTagManager, ColissimoLabel). They are tracked in their own repositories.
- The Flexy `account-order.html.twig` VAT line is wrong for an unrelated reason (`totalAmount - totalAmountWithoutTaxes` mixes discount and postage into the tax figure) and is left alone while the theme rework lands.

Fixes #3668

### DIFF
--- a/core/lib/Thelia/Action/Cart.php
+++ b/core/lib/Thelia/Action/Cart.php
@@ -143,7 +143,7 @@ class Cart extends BaseAction implements EventSubscriberInterface
         try {
             $postage = $this->getPostageByDeliveryModuleId($cart, $dispatcher, $moduleId, $deliveryAddressId);
             $cart
-                ->setPostage((string) ($postage->getAmount() - $postage->getAmountTax()))
+                ->setPostage((string) $postage->getAmount())
                 ->setPostageTax((string) ($postage->getAmountTax() ?? 0.0))
                 ->setPostageTaxRuleTitle($postage->getTaxRuleTitle())
                 ->save();
--- a/core/lib/Thelia/Api/Service/DataAccess/AttributeAccessService.php
+++ b/core/lib/Thelia/Api/Service/DataAccess/AttributeAccessService.php
@@ -238,8 +238,14 @@ class AttributeAccessService
             case 'taxed_postage':
                 $result = $cart->getTaxedPostage();
                 break;
+            case 'untaxed_postage':
+                $result = $cart->getUntaxedPostage();
+                break;
             case 'postage':
                 $result = $cart->getPostage();
+                break;
+            case 'postage_tax':
+                $result = $cart->getPostageTax();
                 break;
             case 'total_price':
             case 'total_price_with_discount':
--- a/core/lib/Thelia/Domain/Checkout/Service/CheckoutPaymentService.php
+++ b/core/lib/Thelia/Domain/Checkout/Service/CheckoutPaymentService.php
@@ -48,7 +48,7 @@ readonly class CheckoutPaymentService
             ->setInvoiceOrderAddressId($invoiceAddressId)
             ->setPaymentModuleId($paymentModuleId)
             ->setDeliveryModuleId($deliveryModuleId)
-            ->setPostage((string) ((float) $cart->getPostage() + (float) $cart->getPostageTax()))
+            ->setPostage((string) $cart->getPostage())
             ->setPostageTax($cart->getPostageTax())
             ->setPostageTaxRuleTitle($cart->getPostageTaxRuleTitle())
             ->setCustomerId($cart->getCustomerId())
--- a/core/lib/Thelia/Domain/Shipping/Service/PostageEstimator.php
+++ b/core/lib/Thelia/Domain/Shipping/Service/PostageEstimator.php
@@ -45,7 +45,7 @@ class PostageEstimator
     }
 
     /**
-     * Return the minimum expected postage for a cart in a given country.
+     * Return the minimum expected postage, tax included, for a cart in a given country.
      *
      * @throws PropelException
      */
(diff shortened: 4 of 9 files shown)


anoziere:
The `test` job is red on this branch for a reason that predates it: `main` is red too.

| ref | run | result |
|---|---|---|
| `2e66295c` (main, before #3692) | [31679909746](https://github.com/thelia/thelia/actions/runs/31679909746) | success |
| `d355bf31` (main, #3692) | [31680172503](https://github.com/thelia/thelia/actions/runs/31680172503) | failure |
| `cf85e5e6` (main HEAD) | [31680530916](https://github.com/thelia/thelia/actions/runs/31680530916) | failure — 539 tests, 12 errors, 1 failure |
| `6636ff8f` (this branch) | [31680424606](https://github.com/thelia/thelia/actions/runs/31680424606) | failure — 541 tests, 12 errors, 1 failure |

Every one of the 12 errors and the single failure is `Call to undefined method: getAnonymizedAt` / `Unknown column AnonymizedAt in model Thelia\Model\Customer`, i.e. the `customer.anonymized_at` column added by #3692 and not yet present in the `thelia/config` release the CI pulls its `schema.xml` from. The failing set is identical on both refs.

This branch adds two tests (539 -> 541) and both pass. `cs-diff`, `phpstan`, and the PHPStan BO Twig job are green.

PR #3704, opened 2026-08-13
fix: move the schema alters out of the released 3.0.0-beta2 update script

anoziere:
`setup/update/sql/3.0.0-beta2.sql` ships in `thelia/setup`, which is tagged
`3.0.0-beta2`. #3692 appended its two `ALTER`s to that script, which is a script
that has already been released.

The update loop stamps `thelia_version` with the name of the last script it ran
and never replays a script at or below that mark. A shop that already ran
`3.0.0-beta2.sql` would therefore never see the appended statements: it would sit
on a `main` build whose code calls `getAnonymizedAt()` against a table that has no
such column, and nothing would tell it why.

So the `ALTER`s move to `setup/update/sql/3.0.0-beta3.sql`, a script no shop has
run yet. `3.0.0-beta2.sql` goes back to exactly the content it had before #3692 —
the diff against its pre-#3692 state is empty.

Checked against a real MariaDB: `setup/thelia.sql` from before #3692 loaded into
one database, then `3.0.0-beta2.sql` and `3.0.0-beta3.sql` applied in order;
`setup/thelia.sql` from `main` loaded into another; the full
`information_schema.COLUMNS` of both compared — identical. `3.0.0-beta3.sql`
replayed cleanly, so a second run is a no-op.


anoziere:
### CI result

`cs-diff` (0 of 1826 files), PHPStan (`[OK] No errors`), the BO Twig PHPStan job
and the unit suite (279 tests) all pass on PHP 8.3 and 8.4.

The integration suite reports the same 12 errors and 1 failure as on `main`
today, identical on both PHP versions, and every one of them is
`Call to undefined method: getAnonymizedAt` / `setAnonymizedAt` or
`Unknown column AnonymizedAt in model Thelia\Model\Customer`. No other exception
or error appears in either job.

Those failures came in with #3692 and clear when `thelia/config` carries
`customer.anonymized_at` and is tagged. This PR only moves SQL statements between
two files, so it neither adds to that set nor can fix it.

PR #3706, opened 2026-08-13
fix: stop offering a hidden state on the address forms

### LINKED ISSUE #3705: A state hidden in the back office is still offered on the address forms
Turning a state off in the back office hides it everywhere except where it matters most: the customer address forms keep offering it.

`AddressCreateForm::getStatesChoices()` (`core/lib/Thelia/Form/AddressCreateForm.php:244-262`) builds the `state` choice list with `StateQuery::create()->filterByCountryId($countryId)` and no filter on `visible`. Three forms inherit that list: `AddressCreateForm`, `AddressUpdateForm` and `CustomerCreateForm`, so the registration form and both address screens are affected.

Everything else honours the flag. The `state` loop defaults to `visible="yes"` (`core/lib/Thelia/Core/Template/Loop/State.php:54`), and the back office shows the state as hidden, so a shop owner has every reason to believe the state is gone.

Concrete case. A shop selling to Italy hides the four Sardinian provinces dissolved in 2016, whose ISO 3166-2 codes were withdrawn (#3690). In the back office the four rows read as not visible. On the front, a customer creating an account or editing an address still finds Carbonia-Iglesias and Olbia-Tempio in the state list and can pick one, so the shop keeps collecting addresses on codes that no longer exist.

The fix is to filter on `visible` where the choices are built.

anoziere:
Turning a state off in the back office hid it everywhere except the customer address forms, which kept offering it.

`AddressCreateForm::getStatesChoices()` built the `state` choice list with no filter on `visible`, and three forms inherit it: `AddressCreateForm`, `AddressUpdateForm` and `CustomerCreateForm`. So registration and both address screens kept listing a state the shop had retired, while the `state` loop, which defaults to `visible="yes"`, and the back office both showed it as gone.

The first commit is the test alone and is expected to fail; the second adds the filter. The test covers the three forms through the shared method.

Fixes #3705

### DIFF
--- a/core/lib/Thelia/Form/AddressCreateForm.php
+++ b/core/lib/Thelia/Form/AddressCreateForm.php
@@ -245,6 +245,7 @@ class AddressCreateForm extends FirewallForm
     {
         $states = StateQuery::create()
             ->filterByCountryId($countryId)
+            ->filterByVisible(1)
             ->useStateI18nQuery()
             ->filterByLocale($this->getLocale())
             ->withColumn(StateI18nTableMap::COL_TITLE, 'title')
--- a/tests/Integration/Form/AddressFormStateChoicesTest.php
+++ b/tests/Integration/Form/AddressFormStateChoicesTest.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Thelia package.
+ * http://www.thelia.net
+ *
+ * (c) OpenStudio <info@thelia.net>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Thelia\Tests\Integration\Form;
+
+use PHPUnit\Framework\Attributes\DataProvider;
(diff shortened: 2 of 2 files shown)


anoziere:
Red then green, on the same branch.

The first commit carries the test alone: run [31680590201](https://github.com/thelia/thelia/actions/runs/31680590201) reports `Tests: 542, Errors: 12, Failures: 4`, the three extra failures being the three data sets of `AddressFormStateChoicesTest` on *a hidden state must not be selectable*. With the filter added, run [31681465683](https://github.com/thelia/thelia/actions/runs/31681465683) reports `Tests: 544, Errors: 12, Failures: 1`, which is the `main` baseline: run [31680937286](https://github.com/thelia/thelia/actions/runs/31680937286) on 1ed2e78 gives `Tests: 541, Errors: 12, Failures: 1`. The remaining 12 errors and 1 failure are `CustomerAnonymizerTest` and `CustomerPurgerTest` calling `getAnonymizedAt()`, waiting on the schema release, and are untouched by this branch.

Issue #3674, opened 2026-08-12
Upgrading a module by zip deletes its row, discarding hooks, positions and configuration


anoziere:
Found while investigating #1145.

`Action\Module::install()` deletes the module row on a zip upgrade and lets `ModuleManagement::updateModule()` recreate it. The module therefore comes back with a **new id**, and everything keyed on the old one is gone: its hooks and their positions, its configuration rows, and any table in another module or theme referencing that id.

An administrator uploading a newer zip of a module they already use expects an upgrade, not a reinstall — nothing in the interface says the configuration will be lost.

Keeping the row in place when the uploaded module has the same namespace would fix this, and it would also make payment and delivery modules upgradable without any schema change, since the deletion guard that refuses to remove a module referenced by an order would no longer be reached. It does change when `destroy()` runs, so it needs its own testing rather than a quick patch.

Same code on both lines; the guard is at `core/lib/Thelia/Action/Module.php:264` on `main` and `:226` on `2.6`.

anoziere:
Fixed on both lines.

- `main`: #3699 — the module row is kept and upgraded in place, plus #3710, which scopes the hook registration to the install path after #3699 had it run on every module refresh.
- `2.6`: #3709 — same behaviour, adapted to the 2.6 files (a merged PR on `2.6` never closes an issue on its own, hence this comment).

An upgrade no longer dispatches `MODULE_DELETE`, so nothing cascades: the hooks and their positions, the module configuration, the hooks the administrator switched off, the module images and title, the profile access rights, the delivery areas and the coupon conditions all survive. The module now receives `update($currentVersion, $newVersion)` instead of a second `install()`.

Three points beyond the ticket:

- The directory used to be wiped by the deletion, so replacing the files was clean by accident. The mirror now runs with `delete` and `override`, otherwise the classes the new version dropped would stay on disk and keep being autoloaded and hooked.
- On `main`, a module installed under `local/modules` was upgraded into `vendor/thelia/modules`, and `Module::getModuleDir()` gives the local copy priority — the old code would have shadowed the new. The files now go where the module already lives.
- An archive that renames the module while keeping its namespace is refused with a message naming both codes, instead of silently creating a second row and a second directory.

Note for the release: `destroy()` no longer runs on an upgrade, `update()` does. A module that relied on destroy+install to migrate has to implement `update()` — the contract `BaseModule` already documents, and what the composer path already did.

[...]

PR #3710, opened 2026-08-13
fix: register a module's hooks on install only, not on every module refresh

anoziere:
Follow-up to #3699.

#3699 made `ModuleManagement::updateModule()` call `registerHooks()` unconditionally, so that reinstalling an archive whose version did not change still registers the hooks the module now declares. That went one step too far: `updateModule()` is also the per-module step of `updateModules()`, which the back-office module list runs on every visit.

`registerHooks()` dispatches `HOOK_CREATE_ALL` or `HOOK_UPDATE` for each hook a module declares, and `Action\Hook` clears the kernel cache on both. Opening the module list therefore wiped the cache — including `var/cache/<env>/sessions` — once per page load, for every module shipping a `getHooks()` declaration.

Found while porting #3699 to 2.6, where the back office renders through the same path and eight functional tests started failing on `file_put_contents(var/cache/test/sessions/...): No such file or directory`.

`updateModule()` now takes `$forceHookRegistration`, which only `Action\Module::install()` sets — the one caller that has just replaced the files on disk. `updateModules()` leaves it off and goes back to registering hooks only when the version actually changed.

`testRefreshingAModuleDoesNotRegisterItsHooksWhenNothingChanged` covers the refresh path; `testReinstallingTheSameVersionRegistersTheHooksTheModuleDeclares`, from #3699, still covers the install path. Both were checked red against the opposite behaviour.

`composer test` 5/5 green (1045 tests), `composer cs-diff` 0/1827, `composer phpstan` 0 errors.

Refs #3674

### DIFF
--- a/core/lib/Thelia/Action/Module.php
+++ b/core/lib/Thelia/Action/Module.php
@@ -400,7 +400,7 @@ class Module extends BaseAction implements EventSubscriberInterface
             $eventDispatcher = $this->container->get('event_dispatcher');
             $moduleManagement = new ModuleManagement($this->container, $eventDispatcher);
             $file = new \SplFileInfo($moduleDescriptorFile);
-            $module = $moduleManagement->updateModule($file, $this->container);
+            $module = $moduleManagement->updateModule($file, $this->container, true);
         } catch (\Throwable $throwable) {
             // The module was deactivated a few lines above so it could be replaced, and the
             // replacement is not going to happen. Put the shop back where it was: a payment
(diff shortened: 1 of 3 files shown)


anoziere:
CI red on a failure this branch does not cause. The integration suite reports 12 errors and 1 failure, all `Unknown column AnonymizedAt in model Thelia\Model\Customer` in `CustomerPurgerTest` and `CustomerAnonymizerTest`.

The same 12 errors and 1 failure run on the HEAD of `main` itself: <https://github.com/thelia/thelia/actions/runs/31681115046> (commit 0c8b469a3, this branch is based on 1ed2e78c1 and shows the identical set). CI rebuilds `local/config/schema.xml` from the released `thelia/config` package, which does not carry `customer.anonymized_at` yet — the column lands with the next config release.

Everything this branch touches is green: `composer test` 5/5 locally (1045 tests), `cs-diff` 0/1827, `phpstan` 0 errors, PHPStan BO Twig green in CI.

PR #3711, opened 2026-08-13
fix: move the state data corrections to the 3.0.0-beta3 update script

anoziere:
#3690 appended its two guarded statements to `3.0.0-beta2.sql`, but `thelia/setup` 3.0.0-beta2 is already tagged. A shop that has run that script will never run it again, so the Ciudad de México rename and the Sardinian province clean-up would only ever reach a fresh install through the seed, never an existing one.

`Update.php` applies every script that sits after the version marker held in the database, so `3.0.0-beta3.sql` is the right place and it already exists (#3704).

`3.0.0-beta2.sql` goes back byte for byte to what it was before #3690; the two blocks move to the end of `3.0.0-beta3.sql`, unchanged. The seed files are untouched.

Refs #3168


anoziere:
The two `CI` jobs are red on the same 12 errors and 1 failure as `main` itself: run [31680937286](https://github.com/thelia/thelia/actions/runs/31680937286) on `main` (1ed2e78) reports `Tests: 541, Assertions: 1292, Errors: 12, Failures: 1`, and this branch reports exactly the same numbers. They all come from `CustomerAnonymizerTest` and `CustomerPurgerTest` calling `getAnonymizedAt()`, which waits on the schema release, and this pull request only moves SQL between two update scripts. `cs-diff` and both PHPStan jobs pass.

Issue #3659, opened 2026-08-12
The registration and activation journey sends two broken mails and generates a route that does not exist


anoziere:
Four defects found while working on #3633, all on the same journey, all reproduced on a fresh install with demo data.

1. **Activation mails ship with an empty subject.** The `message` row `customer_send_code` (id 9) has **zero** `message_i18n` rows, so the database subject is null and the template `email_subject` block is ignored. Verified in the mail catcher: `Subject: ""`. Seed-data defect.
2. **`customer_confirmation` links to a route that does not exist on Thelia 3.** The template points at `/customer/confirm/%token`, which `debug:router` does not know — it only exists in the 2.6 Front module. Every registration sends this dead-link mail *on top of* the activation code mail, so a new customer receives two mails, one of which cannot work.
3. **`customer.create.process` is generated but does not exist.** `thelia-templates/flexy/src/Controller/CustomerController.php:80` builds that route when a visitor submits the login form with "I do not have an account", giving a `RouteNotFoundException`.
4. **An unconfirmed customer has no way back to the activation page.** Login answers "not yet confirmed" and re-sends the dead-link mail; nothing links to the activation page, and registering again is refused. This one needs a product decision rather than a patch — wiring a link is safe since the password is verified first, but where it should lead is a choice.

The first three are defects with obvious fixes. The fourth is the one that leaves a customer stuck, and it is worth deciding explicitly.

anoziere:
Fixed by https://github.com/thelia/thelia/pull/3672 (core, merged first) and https://github.com/thelia-templates/flexy/pull/143 (theme).

**Two of the four points as I wrote them were partly wrong, and the reason is worth recording.** Flexy requires `thelia/front-module`, so the Front module is active on any standard install and *does* serve the two routes I called missing. `debug:router` does not show them because it only lists `router.default` — the known multi-router trap of this project, and I walked straight into it.

- **Point 2**: `/customer/confirm/{token}` exists and works — clicked, the account went to `enable=1`. The real defect is narrower and still real: the core mails a link whose only server is an optional 2.6 module, **on top of** the activation code mail. The core no longer sends it.
- **Point 3**: `customer.create.process` is the POST `/register` route of the Front module. Generating it as a redirect target produces a GET, which the legacy route controller redirects to `/customer/register` **losing the email**. Without the Front module it is a genuine `RouteNotFoundException`. Now the address is carried through and prefilled.
[...]

PR #3712, opened 2026-08-13
fix: move the statements added after the 3.0.0-beta2 tag to the beta3 script

anoziere:
Same trap as #3711, two commits further back. #3660 and #3672 appended to `3.0.0-beta2.sql` on 12 August at 17:07 and 17:38, while `thelia/setup` 3.0.0-beta2 had been tagged that morning at 08:52 UTC. A shop that has run beta2 never runs it again, so neither the Mexican ISO 3166-2 recode nor the activation mail subject would ever reach it.

`3.0.0-beta2.sql` is restored byte for byte to what the tagged package actually ships, checked against `thelia/setup` at the tag rather than against a commit of this repository. The two blocks move into `3.0.0-beta3.sql` with their statements unchanged: 54 lines out, the same 54 lines in.

Order matters inside beta3. The `AGS` to `AGU` and `DIF` to `CMX` recode is placed **before** the Ciudad de México rename moved by #3711, because that rename filters on `state.isocode = 'CMX'`; appended after it, the rename would have found nothing on a shop coming from beta2.

All four statements stay no-ops on a second run: the two recodes match on the outdated value, the `message_i18n` rows are inserted with `INSERT IGNORE`, and the two wording updates are guarded on an empty subject.

Refs #3168 #3659


anoziere:
Identical to the `main` baseline. Run [31681776191](https://github.com/thelia/thelia/actions/runs/31681776191) on `main` (0c8b469) reports `Tests: 544, Assertions: 1298, Errors: 12, Failures: 1`, and this branch reports the same line, with the same 12 errors and 1 failure in `CustomerAnonymizerTest` and `CustomerPurgerTest` on `getAnonymizedAt()`. This pull request moves SQL between two update scripts and touches no PHP. `cs-diff` and both PHPStan jobs pass.

PR #3713, opened 2026-08-13
feat: delete a payment or delivery module an order used, keeping the name it carried

### LINKED ISSUE #1145: Can't delete a Payment/Delivery module referenced by an order.

anoziere:
A payment or delivery module used by at least one order could not be deleted. The two foreign keys on `order` refuse it, and a guard committed the day #1145 was opened turned the constraint violation into an error message telling the administrator to deactivate the module instead. Ten years later the module list of a long-running shop still carries everything it ever tried.

Behind that refusal sits a defect of its own: the order does not conserve the method it used, it points at it. Everything else on an order is a snapshot - `order_address` is a copy, `order_product` freezes the reference, the title and the price, `postage_tax_rule_title` freezes the postage rule name - and the two module ids are the exception. So the name printed on an invoice, mailed in a confirmation or written in an export is resolved from the module row at render time: renaming `Cheque` rewrites every invoice ever issued.

This freezes the name on the order, then lets the module go.

## What happens now when a module is deleted

`Thelia\Action\Module::delete()` writes the name the module carries - in the language of each order - on the orders that used it, sets both references to NULL, and proceeds with the deletion. The foreign keys stay `RESTRICT`: the release is explicit and traceable in the code, not a side effect of a cascade. One update per installed language, no per-order loop.

`order.payment_module_title` and `order.delivery_module_title` stay NULL for as long as the module is installed, since the module row is then the better source. They are mirrored in `order_version` (`order` is versionable) and the migration is guarded, so a second run is a no-op.

## Breaking change: the API can answer null for a module

`Api/Resource/Order.php` declared `public Module $paymentModule` and `public Module $deliveryModule` as non-nullable, with `#[NotBlank]` on the admin write group. Both become `?Module`, defaulting to null, and the two `#[NotBlank]` constraints go: an administrator has to be able to update an order whose module has been deleted.

Consumers of `GET /api/admin/orders/{id}` and `GET /api/front/account/orders/{id}` must accept `paymentModule`/`deliveryModule` being absent, and read `paymentModuleTitle`/`deliveryModuleTitle` - added in the same groups - for the name to display.

That edit is also a fix, not only a contract change. The Propel bridge derives the join type from the nullability of the property (`EagerLoadingExtension`: a non-nullable relation becomes an `INNER JOIN`), so an order whose module id was NULL answered **404 on its own URL** and dropped out of any read that joins the module - silently, since an inner join does not raise, it removes rows. `testOrderWhosePaymentModuleWasDeletedStaysReadable` fails with that 404 on the current code.

## Reading the frozen name

- `OrderExport` keeps its `payment_module_title` / `delivery_module_title` output columns, now `COALESCE(live module title, frozen title)`, so past exports stay comparable.
- The order loop exposes `PAYMENT_MODULE_TITLE` and `DELIVERY_MODULE_TITLE` next to `POSTAGE_TAX_RULE_TITLE`; `PAYMENT_MODULE` and `DELIVERY_MODULE` keep carrying the id, which the invoice and delivery hooks pass on.
- `Order::getPaymentModuleInstance()` and `getDeliveryModuleInstance()` already threw when the module was gone; they now name the order and the module instead of printing an id that no longer resolves.
- The two translation keys of the removed guard are dropped from the nine catalogs that carried them.

## Tests

- `ModuleActionTest::testDeletingAModuleUsedByAnOrderFreezesItsNameOnTheOrder` - installs a module, takes an order through it for both payment and delivery, deletes the module: the row is gone, the order is still there, both ids are NULL and both titles hold the name the merchant had given it. Red on the current code with the guard's `LogicException`.
- `OrderApiTest::testOrderWhosePaymentModuleWasDeletedStaysReadable` - the order stays in the collection, answers 200 on its own URL, and carries the frozen title. Red with a 404 on the current code.

Five suites green (279 unit, 546 integration, 203 api, 12 flexy, 10 back-office), phpstan clean, cs-diff clean, run against a database carrying the two new columns.

## Schema release

The two columns are new, so `thelia/config` and `thelia/setup` need a release carrying them (as `3.0.0-beta4` did for the nullable ids of #3692): CI installs those packages over `local/config/schema.xml` and `setup/`, builds the Propel models from the package copy, and until then `getPaymentModuleTitle()` is undefined and the column is missing from the test database.

Closes #1145


anoziere:
CI is red on one thing only, and it is the schema release, not the code.

```
##[error]Call to an undefined method Thelia\Model\Order::getPaymentModuleTitle().
##[error]Call to an undefined method Thelia\Model\Order::getDeliveryModuleTitle().
 [ERROR] Found 2 errors
```

`thelia/config` is a `thelia-local` package installed over `local/config`, so `composer install` replaces the `schema.xml` of the checkout with the published one and the Propel models are built from that. `3.0.0-beta4` carries the nullable module ids of #3692 but not the two title columns this pull request adds, so the two getters do not exist in the built model. `composer ci` stops at phpstan, which is why the suites did not run - the same signature `main` showed as `Call to undefined method: getAnonymizedAt` between #3692 and the beta4 release. `phpstan-botwig` is green.

Locally, against a database and a `schema.xml` carrying the two columns: five suites green (279 unit, 546 integration, 203 api with the legitimate skip, 12 flexy, 10 back-office), phpstan clean, cs-diff clean on 1828 files.

So this needs a `thelia/config` release carrying `order.payment_module_title` and `order.delivery_module_title` before it can go green, and https://github.com/thelia-templates/default-twig/pull/34 needs both this and that release before it is merged - `phpstan-botwig` clones the theme's `main` and analyses it against the published models.

PR #3720, opened 2026-08-13
feat: store the virtual document of a sale element in its own table

### LINKED ISSUE #3520: Rethink the virtual product / file association (one file per PSE, stored in meta_data)

anoziere:
Closes #3520 (its first question).

The file a virtual product is downloaded from was a `meta_data` row keyed `('virtual', 'pse', sale element id)`, holding the document id as text. Nothing tied it to the two tables it points at:

- deleting the document, or the combination, left the row behind;
- InnoDB does not persist `AUTO_INCREMENT` across a restart, so a sale element created later could be given a freed id and inherit an association nobody set;
- the block declared only a non-unique index, so the "one row per sale element" rule was held by the `findOne()` in `MetaDataQuery::setVal()` alone. Two concurrent writes, or a module inserting a `MetaData` by hand, left two rows and the reader picked one.

`product_sale_elements_virtual_document` gives the association a foreign key on each side and a unique pair, so both deletions cascade and the duplicate cannot exist.

### Why it diverges from the table it mirrors

It is shaped after `product_sale_elements_product_document`, and differs on four points, all deliberate:

| | `product_sale_elements_product_document` | this table |
|---|---|---|
| `position` | absent | `INTEGER NOT NULL DEFAULT 1` |
| `<unique>` | absent | on `(product_sale_elements_id, product_document_id)` |
| `timestampable` | absent | present |
| `isCrossRef` | `true` | not declared |

`timestampable` is what the backfill fills `created_at` with. The unique constraint is exactly what `meta_data` was missing. `position` is what makes several files per combination a second row rather than a second migration — the table allows it, `ProductSaleElements::setVirtualDocument()` keeps writing a single row until that product decision is taken (question 2 of the issue, still open, and it needs an answer on the order side first: `order_product.virtual_document` holds a single file name).

`isCrossRef` is left out because the pair `(product_sale_elements, product_document)` is already bridged by one cross-reference table: a second one makes Propel generate a second, colliding set of `getProductDocuments()` / `addProductDocument()` accessors on `ProductSaleElements`. The table also carries a payload column, so it is not a pure cross-reference.

### Migration

`setup/update/sql/3.0.0-beta3.sql` creates the table, moves the rows, then drops the key — there is no double-writing window to keep, since every reader and writer moves in this PR.

`value` is a `CLOB` any module can write through `MetaDataQuery::setVal()`, so the `INSERT ... SELECT` guards on `is_serialized = 0` and on `value REGEXP '^[0-9]+$'`: MySQL casts `'12-handbook.pdf'` to `12`, which would otherwise become an association with the document the name merely starts with. The rows that cannot be carried over — pointing at a document or a combination that no longer exists, or failing a guard — are listed by a `SELECT` before they go, rather than discarded in silence.

### Callers

- `ProductSaleElements::getVirtualDocuments()` / `getVirtualDocument()`, added in #3687 over `meta_data`, now read the table. Modules that adopted the accessor have nothing to do.
- `ProductSaleElements::setVirtualDocument()` is the matching writer, and the place the single-row rule is enforced.
- `Action\Product::updateVirtualDocumentAssociation()` (#3550) and `Action\ProductSaleElement::cloneVirtualDocumentAssociation()` (#3691) write through it.
- `Order::hasVirtualProductWithDocument()` (#3545) reads `order_product.virtual_document` and is untouched: an order keeps its own copy of the file name, so no order is rewritten.
- No API surface: there is no `MetaData` resource, and the sale element resource never exposed the association.

Out of tree, on the same storage, sent separately: [VirtualProductDelivery](https://github.com/thelia-modules/VirtualProductDelivery/pull/2), [default-twig](https://github.com/thelia-templates/default-twig/pull/35), [back](https://github.com/thelia-templates/back/pull/11).

### Tests

`VirtualDocumentMigrationTest` runs the `INSERT` and `DELETE` read straight out of the shipped update script, so it fails if the script stops doing what it claims: an association is moved with `position` 1 and timestamps, unrelated meta data survives, and the four kinds of row that cannot be carried over are not. Both guards were checked load-bearing — removing either turns two of those tests red.

`ProductSaleElementsVirtualDocumentTest` covers the accessors and both cascades.

### CI

`local/config/schema.xml` gains a table, so `composer install` on `main` extracts the published `thelia/config` over it and PHPStan reports the new classes as unknown until that package is re-tagged. Everything else is green, and the five suites pass locally against the schema carrying the table.

### DIFF
--- a/core/lib/Thelia/Action/Product.php
+++ b/core/lib/Thelia/Action/Product.php
@@ -25,8 +25,6 @@ use Thelia\Core\Event\Feature\FeatureAvDeleteEvent;
 use Thelia\Core\Event\FeatureProduct\FeatureProductDeleteEvent;
 use Thelia\Core\Event\FeatureProduct\FeatureProductUpdateEvent;
 use Thelia\Core\Event\File\FileDeleteEvent;
-use Thelia\Core\Event\MetaData\MetaDataCreateOrUpdateEvent;
-use Thelia\Core\Event\MetaData\MetaDataDeleteEvent;
 use Thelia\Core\Event\Product\ProductAddAccessoryEvent;
 use Thelia\Core\Event\Product\ProductAddCategoryEvent;
 use Thelia\Core\Event\Product\ProductAddContentEvent;
@@ -60,7 +58,6 @@ use Thelia\Model\Map\AttributeTemplateTableMap;
 use Thelia\Model\Map\FeatureTemplateTableMap;
 use Thelia\Model\Map\ProductSaleElementsTableMap;
 use Thelia\Model\Map\ProductTableMap;
-use Thelia\Model\MetaData as MetaDataModel;
 use Thelia\Model\Product as ProductModel;
 use Thelia\Model\ProductAssociatedContent;
 use Thelia\Model\ProductAssociatedContentQuery;
@@ -381,7 +378,6 @@ class Product extends BaseAction implements EventSubscriberInterface
 
     /**
      * Store the document a virtual product is downloaded from, for the default sale element.
-     * The association lives in a meta_data row (virtual key, pse element).
      *
      * A null or negative value means the caller does not handle the association (a product with
      * several combinations sets it per combination), 0 removes it, and a document id sets it.
@@ -403,19 +399,7 @@ class Product extends BaseAction implements EventSubscriberInterface
             return;
         }
 
-        if (0 === (int) $virtualDocumentId) {
-            $this->eventDispatcher->dispatch(
-                new MetaDataDeleteEvent('virtual', MetaDataModel::PSE_KEY, $defaultPse->getId()),
-                TheliaEvents::META_DATA_DELETE
-            );
-
-            return;
-        }
-
-        $this->eventDispatcher->dispatch(
-            new MetaDataCreateOrUpdateEvent('virtual', MetaDataModel::PSE_KEY, $defaultPse->getId(), (int) $virtualDocumentId),
-            TheliaEvents::META_DATA_UPDATE
-        );
+        $defaultPse->setVirtualDocument(0 === (int) $virtualDocumentId ? null : (int) $virtualDocumentId);
     }
 
     public function updateSeo(UpdateSeoEvent $event, $eventName, EventDispatcherInterface $dispatcher): mixed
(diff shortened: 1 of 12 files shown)


anoziere:
### CI, single cause

Both matrix jobs stop at PHPStan with three errors, and all three are the three classes this PR adds:

```
  core/lib/Thelia/Model/ProductSaleElements.php:167
 [ERROR] Found 3 errors
```

The Base classes come from Propel, generated out of `local/config/schema.xml`. The `composer install` of the CI extracts the published `thelia/config` package over that file, and the published package does not carry the new table yet, so the generator never produces them. The signature therefore cannot appear on `main` — and does not, the run on the commit this branch sits on is green. It clears itself the moment `thelia/config` is re-tagged, which is batched with the rest of the schema work.

Nothing else is red:

- `cs-diff`: `Found 0 of 1837 files that can be fixed`, on both 8.3 and 8.4.
- `PHPStan level 5 (BO Twig bundle)`: pass.
- the five PHPUnit suites never run here, since `composer ci` stops at PHPStan. They were run locally against a database and a Propel build made from the schema carrying the table, on the rebased branch: unit 284, integration 571, api 203 (1 skip), http-flexy 12 (2 skips), http-backoffice 10 (2 deprecations), no failure. Same skips and deprecations as `main`, plus the nine tests this PR adds. `cs-diff` and PHPStan are also green locally, where the schema does carry the table.

anoziere:
Rebased on 9e29412c2. Re-run locally against the schema carrying the table, on that base: unit 284, integration 573, api 203 (1 skip), http-flexy 12 (2 skips), http-backoffice 10 (2 deprecations), no failure; `cs-diff` 0 of 1843; PHPStan and the BO Twig level-5 gate both clean.

The CI now reports five PHPStan errors rather than three: two of them are `OrderPostageTax` from #3721, already on `main` and red there for the same reason — the published `thelia/config` does not carry either new table yet. Both clear on the next `thelia/config` tag.

PR #3721, opened 2026-08-13
feat: split the postage tax between the rates a mixed cart carries

### LINKED ISSUE #3701: Postage VAT breakdown for mixed-rate carts
Split out of #1664, whose title ("More tax rules for postage") is satisfied by #3664 and the back-office screen that came with #3695. What that issue really carried, and what remains, is the tax breakdown of postage in a mixed-rate cart.

## The problem

`order.postage_tax` is a single blended figure. It cannot express the European rule that postage follows the goods: a cart holding a 20% book-case and a 5.5% book pays one postage, and today the whole of its tax is attributed to one rule (`order.postage_tax_rule_title`). Nothing records how the postage tax splits.

Default behaviour must not change on upgrade: a minor version cannot silently move the VAT a shop declares.

## Where the calculation goes

As a post-processing step on the `OrderPostage` value object, at the two points where the cart and the postage already meet:

- `core/lib/Thelia/Action/Cart.php:631-639` — the `DeliveryPostageEvent` carries the `Cart` (`core/lib/Thelia/Core/Event/Delivery/DeliveryPostageEvent.php:43`, getter `:50`) and receives the `OrderPostage` back
- `core/lib/Thelia/Domain/Shipping/Service/PostageEstimator.php:75` — the estimation path

This keeps `buildOrderPostage()`'s signature intact, so no delivery module in the ecosystem is affected, and it keeps the estimate and the order consistent — one place to fix. Widening the signature, or letting each carrier split its own postage, both cost more and buy nothing.

`core/lib/Thelia/Model/OrderPostage.php:22` is a plain value object, not a Propel model: it can carry the breakdown as an extra collection, its three current fields untouched, with no schema change until the order is persisted.

## Machinery already available

`core/lib/Thelia/Domain/Taxation/TaxEngine/TaxCalculatorInterface.php:41` and `:45` take an `OrderProductTaxCollection &$taxCollection` by reference and fill it with one line per tax — the same mechanism that feeds `order_product_tax`. A postage breakdown can reuse it as is, one call per rate group.

Do **not** reuse `Calculator::computeCartTaxFactor()` (`core/lib/Thelia/Domain/Taxation/TaxEngine/Calculator.php:173`): it averages the per-line factors unweighted (`array_sum(...) / count(...)`), which is not a valid basis for a value pro rata.

## Persistence

A child table, to be batched with the other pending schema changes rather than pulling a `thelia/config` release of its own:

```xml
<table name="order_postage_tax" namespace="Thelia\Model">
  <column autoIncrement="true" name="id" primaryKey="true" required="true" type="INTEGER" />
  <column name="order_id" required="true" type="INTEGER" />
  <column name="title" required="true" size="255" type="VARCHAR" />
  <column name="description" type="CLOB" />
  <column defaultValue="0.000000" name="untaxed_amount" required="true" scale="6" size="16" type="DECIMAL" />
  <column defaultValue="0.000000" name="amount" required="true" scale="6" size="16" type="DECIMAL" />
  <foreign-key foreignTable="order" name="fk_order_postage_tax_order_id" onDelete="CASCADE" onUpdate="RESTRICT">
    <reference foreign="id" local="order_id" />
  </foreign-key>
  <index name="idx_order_postage_tax_order_id"><index-column name="order_id" /></index>
  <behavior name="timestampable" />
</table>
```

`untaxed_amount` is deliberate and is **not** a mirror of `order_product_tax`, which carries `amount` + `promo_amount` and no untaxed column (`local/config/schema.xml:1497-1502`). Without it the rate cannot be reconstructed from the row. No foreign key to `tax_rule`: same snapshot discipline as `order_product_tax`.

Invariants to test: `sum(untaxed_amount) + sum(amount) == order.postage` and `sum(amount) == order.postage_tax`, with the rounding remainder given to the largest share.

Written at order creation, `core/lib/Thelia/Domain/Checkout/Service/CheckoutPaymentService.php:51-53`, after the order's own `save()`. The cart stays flat: `cart.postage*` is an estimate recomputed on every change, no reason to double the writes.

## Strategy

A `Config` variable, `postage_tax_strategy`, with `single_rule` / `pro_rata` / `highest_rate`. Default `single_rule`, byte-for-byte today's behaviour; pro rata is announced in the release notes and ticked by the merchant. No column.

Existing orders need no backfill: the flat columns remain the reference, and an order with no lines reads correctly as "one rate, named in `postage_tax_rule_title`". No recomputing of a paid invoice.

## Readers to update once lines exist

- coupon: `core/lib/Thelia/Action/Coupon.php:243-245` and `:258-260` null the three postage columns for free shipping — they must clear the breakdown too
- totals: `core/lib/Thelia/Model/Order.php:279-280` and `:332-333`; `getUntaxedPostage()` at `:364-366`
- PDF invoice: `thelia-templates/pdf`, `invoice.html.twig:208-210`, `:227`, `:231`, `:247` (already prints untaxed postage plus tax)
- back-office order detail: `OrderDetailContextBuilder.php:68-70`, `order/detail.html.twig:228-231`
- accounting export: `core/lib/Thelia/Domain/DataTransfer/Export/Type/OrderExport.php:71-72` (raw SQL)
- API: `core/lib/Thelia/Api/Resource/Order.php:433-462`

Nothing changes for them while an order has no lines. Rendering the breakdown is worth a separate batch from the calculation itself.

## Ordering

#3668 covers the opposite untaxed/taxed conventions of `cart.postage` (`core/lib/Thelia/Action/Cart.php:146-148`, untaxed) and `order.postage` (`CheckoutPaymentService.php:51-53`, taxed). The split computes untaxed and persists on the order: if #3668 renames anything, land it first so this is written against the final names.

anoziere:
`order.postage_tax` is a single blended figure and `postage_tax_rule_title` names one rule. A cart holding a book-case at 20 % and a book at 5.5 % pays one postage, and today the whole of its tax is attributed to one rule — nothing records how it splits, so the European rule that the postage follows the goods cannot be expressed.

Default behaviour does not change. A minor version cannot silently move the VAT a shop declares, so the split is a setting the merchant ticks.

## Where the calculation goes

A post-processing step on the `OrderPostage` value object, at the two points where the cart and the postage already meet:

- `Action\Cart::getPostageByDeliveryModuleId()` — the `DeliveryPostageEvent` already carries the `Cart` and gives the `OrderPostage` back
- `PostageEstimator::computePostageForModule()` — the estimation path, so an estimate and the cart it turns into never announce two different taxes

`buildOrderPostage()` keeps its signature, so **no delivery module in the ecosystem is affected**, and the estimate and the order stay consistent with one place to fix.

## The setting

`postage_tax_strategy`, a `Config` variable, no column:

| Value | What it does |
|---|---|
| `single_rule` *(default)* | Today's behaviour to the byte: one rule for the whole postage, no breakdown written |
| `pro_rata` | The untaxed postage is spread over the untaxed value each rate accounts for |
| `highest_rate` | The whole postage follows the dearest rate the cart carries |

`Calculator::computeCartTaxFactor()` is deliberately **not** reused: it averages the per-line factors unweighted, which is not a valid basis for a value pro rata. Each rate group goes through `TaxCalculatorInterface::getTaxAmountFromUntaxedPrice()` on a calculator loaded with its own rule.

The untaxed postage is the fixed part — it is the price the delivery module quoted and what the merchant charges. The tax is the derived figure, so opting in recomputes the tax and the taxed total from it. On a 10.00 postage with 200.00 at 20 % and 100.00 at 5.5 %, `pro_rata` gives 6.67 at 20 % and 3.33 at 5.5 %, so 1.51 of tax instead of 2.00.

## Persistence

`order_postage_tax`, one row per rule, written at order creation inside the order transaction (`OrderFacade::createOrder()`, so the admin-created order goes through it too).

`untaxed_amount` is deliberate and is **not** a mirror of `order_product_tax`, which carries `amount` + `promo_amount` and no untaxed column: without it the rate cannot be reconstructed from the row. No foreign key to `tax_rule` — the same snapshot discipline `postage_tax_rule_title` already follows.

One row per **rate group**, not per tax: `untaxed_amount` is a base, and a base repeated across the taxes of one rule would break the sums. The rounding remainder of each column goes to the largest line, so the two invariants hold exactly:

```
sum(untaxed_amount) + sum(amount) == order.postage
sum(amount)                       == order.postage_tax
```

The order columns are what the customer was charged, so they are the reference: the lines are reconciled against them, never the other way round.

Existing orders need no backfill. An order with no line reads correctly as one rate named in `postage_tax_rule_title` — which is also what every order placed under the default strategy keeps looking like. No recomputing of a paid invoice.

A free-shipping coupon needs nothing either: it nulls the cart postage columns, the order is then placed with a postage of zero and no line is written.

## Tests

`tests/Integration/Domain/Shipping/PostageTaxBreakdownTest.php` drives the real `Action\Cart` post-processing on a mixed-rate cart, replacing only the delivery module's own answer, then takes the cart through a real `CheckoutPaymentService::pay()`:

- the default strategy leaves the quote untouched and writes no row
- `pro_rata` splits 6.67 / 3.33 with the remainder on the largest share, and the exact taxes 1.33 / 0.18
- `highest_rate` taxes the whole postage at 20 %
- a free postage is not split
- the placed order carries two rows whose sums match its own columns

Verified red first: with the split neutralised, the three cases that need it fail (`0` lines against `2`), and the three that assert the unchanged default stay green.

Five suites green locally against a database carrying the new table (284 unit, 562 integration, 203 api, 12 flexy, 10 back-office), `phpstan` and `phpstan-botwig` clean, `cs-diff` clean.

## Schema release

The table is new, so `thelia/config` and `thelia/setup` need a release carrying it: CI installs those packages over `local/config/schema.xml` and `setup/`, builds the Propel models from the package copy, and until then `Thelia\Model\Base\OrderPostageTax` does not exist and the table is missing from the test database. The migration is appended to `setup/update/sql/3.0.0-beta3.sql`, guarded with `CREATE TABLE IF NOT EXISTS`, so a second run is a no-op.

## Not in this PR

- **Rendering the breakdown** — PDF invoice, back-office order detail, accounting export and the API resource. All of them read correctly today while an order has no line, and showing the detail is worth its own batch.
- **A back-office screen for `postage_tax_strategy`.** The Twig back office lives in its own repository; without a screen the setting is reachable only through `config`, exactly the gap #3664 left behind. It should land next.

Fixes #3701

### DIFF
--- a/core/lib/Thelia/Action/Cart.php
+++ b/core/lib/Thelia/Action/Cart.php
@@ -32,6 +32,7 @@ use Thelia\Core\HttpFoundation\Session\Session;
 use Thelia\Core\Security\SecurityContext;
 use Thelia\Domain\Cart\Exception\NotEnoughStockException;
 use Thelia\Domain\Cart\Service\CartAddressService;
+use Thelia\Domain\Shipping\Service\PostageTaxBreakdownCalculator;
 use Thelia\Model\AddressQuery;
 use Thelia\Model\Base\CustomerQuery;
 use Thelia\Model\Base\ProductSaleElementsQuery;
@@ -65,6 +66,7 @@ class Cart extends BaseAction implements EventSubscriberInterface
         protected SecurityContext $securityContext,
         protected ContainerInterface $container,
         protected CartAddressService $cartAddressService,
+        protected PostageTaxBreakdownCalculator $postageTaxBreakdownCalculator,
     ) {
     }
 
@@ -638,11 +640,19 @@ class Cart extends BaseAction implements EventSubscriberInterface
             TheliaEvents::MODULE_DELIVERY_GET_POSTAGE
         );
 
-        if ($deliveryPostageEvent->getPostage()) {
-            return $deliveryPostageEvent->getPostage();
+        $postage = $deliveryPostageEvent->getPostage();
+
+        if (!$postage instanceof OrderPostage) {
+            return new OrderPostage();
         }
 
-        return new OrderPostage();
+        // The delivery module quotes a postage under a single tax rule; whether
+        // that tax follows the goods instead is a shop decision, and the cart is
+        // only known here. Doing it on the way out keeps buildOrderPostage()'s
+        // signature intact, so no delivery module has to change.
+        $this->postageTaxBreakdownCalculator->applyToPostage($postage, $cart, $country, $state);
+
+        return $postage;
     }
 
     protected function dispatchNewCart(EventDispatcherInterface $dispatcher): CartModel
--- a/core/lib/Thelia/Domain/Order/OrderFacade.php
+++ b/core/lib/Thelia/Domain/Order/OrderFacade.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 
 namespace Thelia\Domain\Order;
 
+use Propel\Runtime\Connection\ConnectionInterface;
 use Propel\Runtime\Exception\PropelException;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Thelia\Core\Security\User\UserInterface;
@@ -27,12 +28,16 @@ use Thelia\Domain\Order\Service\StockPolicy;
 use Thelia\Domain\Order\Service\TaxProvider;
 use Thelia\Domain\Order\Service\TranslationProvider;
 use Thelia\Domain\Order\Service\VirtualProductHandler;
+use Thelia\Domain\Shipping\Service\PostageTaxBreakdownCalculator;
 use Thelia\Exception\TheliaProcessException;
 use Thelia\Model\Cart as CartModel;
 use Thelia\Model\ConfigQuery;
+use Thelia\Model\Country;
 use Thelia\Model\Currency as CurrencyModel;
 use Thelia\Model\Lang as LangModel;
 use Thelia\Model\Order as ModelOrder;
+use Thelia\Model\OrderAddressQuery;
+use Thelia\Model\OrderPostageTax;
 use Thelia\Model\OrderProductTax;
 use Thelia\Model\OrderStatusQuery;
 
@@ -49,6 +54,7 @@ readonly class OrderFacade
         private OrderProductFactory $orderProductFactory,
         private OrderRefGeneratorInterface $orderRefGenerator,
         private StockDecrementer $stockDecrementer,
+        private PostageTaxBreakdownCalculator $postageTaxBreakdownCalculator,
     ) {
     }
 
@@ -99,6 +105,8 @@ readonly class OrderFacade
 
             $placedOrder->setStatusId(OrderStatusQuery::getNotPaidStatus()?->getId());
             $placedOrder->save($connection);
+
+            $this->persistPostageTaxBreakdown($placedOrder, $cart, $taxCountry, $lang, $connection);
 
             $manageStockOnCreation = $placedOrder->isStockManagedOnOrderCreation($dispatcher);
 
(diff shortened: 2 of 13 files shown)


anoziere:
## CI failure is the schema release cycle, and nothing else

Both `CI (8.3)` and `CI (8.4)` stop on exactly two PHPStan errors, and they are the same two:

```
  core/lib/Thelia/Model/OrderPostageTax.php:19
  core/lib/Thelia/Model/OrderPostageTaxQuery.php:19
[ERROR] Found 2 errors
```

The full failed-step log of run [31689728708](https://github.com/thelia/thelia/actions/runs/31689728708), deduplicated, contains nothing else:

```
[ERROR] Found 2 errors
Process completed with exit code 1.
```

`cs-diff` runs before PHPStan in the `ci` script and passed on both versions. `PHPStan level 5 (BO Twig bundle)` is green.

**Cause.** The workflow installs `thelia/config` over `local/config/schema.xml` before the Propel models are built, and the published package does not carry the new table — `repos/thelia/config/contents/schema.xml` at `7d350b9` has zero occurrences of `order_postage_tax`. The two `Base/` classes therefore do not exist for the analysis step, and the five suites never run because the script stops there. Same cycle as #3692 and #3713: this clears once `thelia/config` and `thelia/setup` carry the table and are tagged, with no change to this branch.

**What was run locally instead**, against a database built from the complete `local/config/schema.xml` of this branch:

| | |
|---|---|
| `phpstan` | No errors |
| `phpstan-botwig` | No errors |
| `cs-diff` | 0 of 1835 |
| unit | 284 |
| integration | 562 |
| api | 203 (1 skip) |
| http-flexy | 12 (2 skips) |
| http-backoffice | 10 |

Merging with `--admin` on that basis.

PR #3723, opened 2026-08-13
fix: let a country carry states without requiring one

### LINKED ISSUE #3696: country.has_states cannot express an optional state
The `country.has_states` column is a boolean, and every consumer reads it as "this country requires a state":

- `AddressCountryValidationTrait::verifyState()` only inspects the submitted state when `Country::getHasStates()` is true, and then makes it mandatory: *You should select a state for this country.*
- The `country` loop exposes `$HAS_STATES` and filters on it, so a template uses it to decide whether the state field is shown at all.

There is no value for "this country has states, and picking one is optional".

The seed shows what that costs. #3536 added the 101 French departments as states of France, but France is seeded `has_states = 0` (8th column of the `FRA` row in `setup/insert.sql`), so no address form offers them: the rows are in the database and unreachable. Setting the flag to 1 instead would make the department mandatory on every French address, which is not what a French shop expects.

A third value (none / optional / required), or a separate column, would let a country carry states without forcing one on the customer.

Found while closing #3168.

anoziere:
Fixes #3696

`country.has_states` carried two meanings at once: whether the country has states at all, and whether picking one is mandatory. Nothing could express "states exist, choosing one is optional", so the 101 French departments added by #3536 sat in the database unreachable — France is seeded `has_states = 0`, and setting the flag would have made the department mandatory on every French address.

The two questions are now answered by two different things:

- **what a form offers** comes from the visible `state` rows of the country. This is already how `AddressCreateForm::getStatesChoices()` works, so the departments become selectable with no data change;
- **whether the field may be left empty** stays `has_states`, unchanged for the United States, Canada, Italy, Japan, Mexico, Argentina and Indonesia.

No schema change, no seed change, no migration: France keeps `has_states = 0` and gains an optional department.

### What changes

- `Country::hasSelectableStates()` — the country carries at least one visible state.
- `AddressCountryValidationTrait::verifyState()` — a submitted state is checked against its country whether or not the country requires one, so an optional department no longer survives a change of country in the same submit. A state is required only when the flag is set **and** the country carries some: a country flagged as requiring a state but holding none used to reject every address with an error the customer could not act on.
- `Api\Resource\Address::verifyState()` — same rule, and it no longer fatals: posting an address without a state to a country that requires one answered `500` (`Call to a member function getPropelModel() on null`) instead of a validation error. The callback now also runs on front writes, which had no state validation at all.
- `Tools\AddressFormat` — a state stored on an address is rendered even when the country leaves the choice optional; it used to be dropped.
- `Loop\Country` — new `HAS_SELECTABLE_STATES` output, resolved in one query for the whole loop. `HAS_STATES` keeps its meaning, now documented as "a state is required", so a template can tell the two apart.

### Tests

Each was checked failing first:

- `AddressStateValidationTest` — an out-of-country state is rejected on an optional country; an empty selection is still rejected when required; a country requiring a state but holding none does not block the form.
- `AddressFormatStateTest` — the state survives on the formatted address of a country that does not require one.
- `AddressApiTest` (admin and front) — `422` instead of `500` on a missing required state, and the out-of-country state is rejected on both sides.
- `CountryLoopStatesTest` — France reports `HAS_STATES = 0` and `HAS_SELECTABLE_STATES = 1`.

### Behaviour change to note

Front API address writes (`/api/front/account/addresses`) now validate the state. A client sending a state that belongs to another country, or omitting one for a country that requires it, gets a `422` where the address used to be saved as is.

### DIFF
--- a/core/lib/Thelia/Api/Resource/Address.php
+++ b/core/lib/Thelia/Api/Resource/Address.php
@@ -458,33 +458,45 @@ class Address implements PropelResourceInterface
         }
     }
 
-    #[Callback(groups: [self::GROUP_ADMIN_WRITE, Customer::GROUP_ADMIN_WRITE])]
+    #[Callback(groups: [self::GROUP_ADMIN_WRITE, self::GROUP_FRONT_WRITE, Customer::GROUP_ADMIN_WRITE])]
     public function verifyState(ExecutionContextInterface $context): void
     {
         $resource = $context->getRoot();
 
-        if (isset($resource->country) && null !== ($country = $resource->getCountry()->getPropelModel()) && $country->getHasStates()) {
-            if (null !== $state = $resource->getState()->getPropelModel()) {
-                if ($state->getCountryId() !== $country->getId()) {
-                    $context->addViolation(
-                        Translator::getInstance()->trans(
-                            "This state doesn't belong to this country.",
-                            [],
-                            null,
-                            'en_US',
-                        ),
-                    );
-                }
-            } else {
+        if (!isset($resource->country) || null === $country = $resource->getCountry()?->getPropelModel()) {
+            return;
+        }
+
+        $state = $resource->getState()?->getPropelModel();
+
+        // A state stays tied to its country whether or not the country requires one:
+        // an optional department must not be kept when the address moves elsewhere.
+        if (null !== $state) {
+            if ($state->getCountryId() !== $country->getId()) {
                 $context->addViolation(
                     Translator::getInstance()->trans(
-                        'You should select a state for this country.',
+                        "This state doesn't belong to this country.",
                         [],
                         null,
                         'en_US',
                     ),
                 );
             }
+
+            return;
+        }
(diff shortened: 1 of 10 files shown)


anoziere:
CI is red on the two PHPStan errors `main` already carries — `Thelia\Model\OrderPostageTax` and `OrderPostageTaxQuery` extend Base classes that cannot be generated, because `local/config/schema.xml` has no `order_postage_tax` table yet. Same failure on main: https://github.com/thelia/thelia/actions/runs/31690677335. Neither file belongs to this branch; they come in through the merge commit CI builds. PHPStan runs before the suites in the `ci` script, so the five PHPUnit suites never started here either.

Run locally on this branch against a freshly prepared database: `cs-diff` 0/1833, `phpstan` 0 errors, unit 279 green, api 206 green, integration 553/557 — the four remaining come from `GaplessSequenceGeneratorTest`, which reaches the development database in that sandbox.

Issue #3725, opened 2026-08-13
generate:sql cannot run on Thelia 3, so insert.sql.tpl has drifted from insert.sql


anoziere:
`setup/insert.sql` is generated from `setup/insert.sql.tpl` by `generate:sql`. That command cannot run on Thelia 3:

```
$ php Thelia generate:sql
The "thelia.parser" service or alias has been removed or inlined when the container was compiled.
  at GenerateSQLCommand.php:97
```

Making the service public would not be enough. `insert.sql.tpl` is a Smarty template — `{foreach}`, `{if}`, `{intl}` — and the command drives it through the Smarty API:

```php
$this->parser->unregisterPlugin('function', 'intl');
$this->parser->registerPlugin('function', 'intl', $this->translate(...));
```

Core no longer ships a Smarty parser, requires no `smarty/smarty`, and `FallbackParserPass` registers a `ParserFallback` when no bundle provides `thelia.parser`. So the only thing left under `thelia.parser` in a stock install cannot render the template.

`insert.sql` has therefore been maintained by hand, and it no longer matches what the template describes — not only in placeholder syntax (#3724) but in content:

| table | in the template only | in insert.sql only |
| --- | --- | --- |
| `message_i18n` | — | 9 |
| `resource_i18n` | 48 | — |
| `hook_i18n` | 1316, 1317, 1318 | 1425 |

`insert.sql.tpl` is documentation rather than a source today, and nothing can check the shipped seed against it. Either the generator moves to a parser core still has, or the template goes and `insert.sql` becomes the source it already is in practice.

Found while closing #3697.

Issue #3697, opened 2026-08-13
The seed only carries 7 of the 18 languages shipped in setup/I18n


anoziere:
`setup/I18n/` ships 18 translation files. `setup/insert.sql` only carries rows for the 7 locales seeded in the `lang` table: `cs_CZ`, `de_DE`, `en_US`, `es_ES`, `fr_FR`, `it_IT`, `ru_RU`.

The other 11 (`ar_SA`, `el_GR`, `fa_IR`, `hu_HU`, `id_ID`, `pl_PL`, `pt_BR`, `pt_PT`, `sk_SK`, `tr_TR`, `uk_UA`) are translated in the repository and never reach a database. Adding one of them in the back office gives a shop with no country title, no state title and no message title in that language, although the wording sits in `setup/I18n/<locale>.php`. Nothing replays those translations for the rows already installed.

`generate:sql` is not the obstacle: it walks every file of `setup/I18n` unless `--locales` restricts it. What is missing is a path from those files to an existing database when a language is added.

Found while closing #3168, where it explains why `state_i18n` covers 7 locales for 413 states.

PR #3726, opened 2026-08-13
fix: bring the seed sources to the Twig syntax the mailer renders

### LINKED ISSUE #3724: The seed sources still carry Smarty placeholders while insert.sql carries Twig
The mail templates shipped in `templates/email/default` are all Twig, and `Message::buildMessage()` renders the `message_i18n` subject through the same parser. `setup/insert.sql` was brought to that syntax:

```
(1, 'en_US', 'Order confirmation sent to the customer', 'Your order {{ order_ref }} at {{ config(\"store_name\") }}', NULL, NULL),
```

The sources it is generated from were not. `setup/insert.sql.tpl` still asks for the Smarty wording, and the eighteen files of `setup/I18n` still key and translate it:

```
# setup/insert.sql.tpl
(1, '{$locale}', {intl l='Order confirmation sent to the customer' locale=$locale}, {intl l='Your order {$order_ref} at {config key="store_name"}' locale=$locale}, NULL, NULL),

# setup/I18n/fr_FR.php
'Your order {$order_ref} at {config key="store_name"}' => 'Votre commande {$order_ref} chez {config key="store_name"}',
```

Three consequences.

A language added after the install now takes its wording from `setup/I18n` (#3697). Its mail subjects arrive in Smarty, and the Twig mail parser prints `{$order_ref}` and `{config key="store_name"}` to the customer. The subject also fails to match its key, since the key recovery goes through the `en_US` wording that `insert.sql` stores in Twig, so most subjects are not translated at all.

A `generate:sql` would take `insert.sql` back to Smarty for every locale, undoing the hand conversion.

Two translators broke the tag itself and nothing caught it, because no seed source is checked against the syntax the mailer expects:

```
# setup/I18n/tr_TR.php
'Your order {$order_ref} at {config key="store_name"}' => 'Sipariş {$order_ref}, {config anahtarı "store_name" =}',

# setup/I18n/sk_SK.php
'A {config key="store_name"} account has been created for you' => '{kľúč config = "store_name"} účet bol vytvorený pre vás',
```

Separately, `%store` is a Symfony parameter and the seed passes no parameters, so it reaches the database as it stands. `insert.sql` ships it to every Russian shop:

```
(6, 'ru_RU', 'Письмо отсылаемое администратору при запросе нового пароля', 'Ваш новый пароль для %store', NULL, NULL),
```

Found while closing #3697.

anoziere:
Fixes #3724

The mail templates are Twig and `Message::buildMessage()` renders a `message_i18n` subject through the same parser. `setup/insert.sql` had been brought to that syntax by hand; `setup/insert.sql.tpl` and the eighteen files of `setup/I18n` had not.

Since #3697 a language added after the install takes its wording from `setup/I18n`, so this reached customers: the subject arrived in Smarty for the mailer to print verbatim, and most subjects were not translated at all — the key recovery goes through the `en_US` wording, which `insert.sql` stores in Twig, so it matched nothing.

## What changes

`{$order_ref}` → `{{ order_ref }}`, `{$password}` → `{{ password }}` and `{config key="store_name"}` → `{{ config("store_name") }}`, in the `l='…'` keys of `insert.sql.tpl` and in the keys and translations of `setup/I18n`. Keys and translations move together so the catalogs stay aligned with the template.

Two translators had broken the tag itself — `{config anahtarı "store_name" =}` in `tr_TR`, `{kľúč config = "store_name"}` in `sk_SK` — and those subjects were unusable whatever the parser. They now carry the same call as the rest.

`%store` is a Symfony parameter and the seed passes no parameters, so it was stored as it stands. Three translations resolved to it, and `insert.sql` shipped it to every Russian shop; `setup/update/sql/3.0.0-beta3.sql` carries the `REPLACE` for shops already installed. `%store` stays where it only keys a translation.

The dead `store={config key="store_name"}` attribute goes from the `intl` tag of message 6: `GenerateSQLCommand::translate()` passes no parameters, so it never did anything, and converting it would have left the Smarty tag malformed.

## Tests

`generate:sql` cannot run on Thelia 3 (#3725), so regenerating `insert.sql` and comparing is not available. `tests/Unit/Install/I18n/SeedPlaceholderSyntaxTest.php` stands in for it and is stronger on the part that matters: for the seven shipped locales it resolves every `l='…'` key of the `message_i18n` block through `setup/I18n/<locale>.php` and asserts the result is character for character what `insert.sql` ships — 30 000 assertions covering the whole catalogs. It also refuses a Smarty placeholder in any seed source and a translation resolving to `%store`.

Reverting one catalog entry to its Smarty wording turns both guards red, so the test can fail:

```
fr_FR.php still translates to a Smarty placeholder.
insert.sql message 1 in fr_FR has a subject the seed sources do not produce.
-'Votre commande {$order_ref} chez {config key="store_name"}'
+'Votre commande {{ order_ref }} chez {{ config("store_name") }}'
```

`tests/Integration/Action/LangSeedI18nTest.php` now asserts a language added after the install receives `Twoje zamówienie {{ order_ref }} w {{ config("store_name") }}` — translated, and in the syntax the mailer renders. Before this change that row fell back to the English wording.

## Left for #3725

`generate:sql` is unrunnable and `insert.sql.tpl` has drifted from `insert.sql` in content too (message 9, resource 48, hooks 1316-1318 against 1425). This PR only aligns the placeholder syntax; it does not make the template a source again.


anoziere:
Merging with `--admin`: the red `test` check is the known `thelia/config` staleness, not this branch.

The whole failure is PHPStan missing generated Propel Base classes for two tables merged earlier in this wave:

```
```

`order_postage_tax` comes from #3721 and `product_sale_elements_virtual_document` from #3520; both wait on the `thelia/config` release that closes the wave. The same run is red on `main` itself and on every open PR.

This branch cannot produce those errors: it changes `setup/` and `tests/` plus one comment line, and PHPStan analyses `core` only. `cs-diff` passes in the same run (0 of 1848). PHPStan gates before PHPUnit, so no suite ran — locally the 9 unit and 6 integration tests of this change are green, and reverting one `fr_FR` entry to its Smarty wording turns both new guards red.

PR #3727, opened 2026-08-13
test: cover the postage tax strategy screen

### LINKED ISSUE #3722: Postage tax strategy has no back-office screen
#3721 gave shops a choice of how the postage tax splits across a mixed-rate cart, through the `postage_tax_strategy` variable: `single_rule` (the default, today's behaviour), `pro_rata` and `highest_rate`. Nothing in the back office writes it, so the two opt-in strategies cannot be turned on. Same hole as #3695, which #3664 had left on the per-module postage tax rule.

### What the code shows

`postage_tax_strategy` appears in three places on `main`:

- `core/lib/Thelia/Domain/Shipping/Enum/PostageTaxStrategy.php:38` — the key, read at `:45`
- `core/lib/Thelia/Domain/Shipping/Service/PostageTaxBreakdownCalculator.php:128` — the only reader
- `tests/Integration/Domain/Shipping/PostageTaxBreakdownTest.php` — the tests that write it

No controller, no form, no template, and no seed row in `setup/insert.sql`. The variables editor can create the row by hand, but that is a raw key/value field: no wording, no list of the accepted values, and a typo falls back to `single_rule` without saying so (`PostageTaxStrategy::fromShopConfiguration()`).

### Expected

The strategy is picked from the tax rules page, next to the shop-wide delivery tax rule and the per-module rules added in #3695, with each option named and explained, and a word on the consequence: the two other strategies move the VAT the shop declares on mixed carts.

No schema change: `config` already holds the value.

anoziere:
Closes #3722.

#3721 gave shops a choice of how the postage tax splits across a mixed-rate cart, and nothing wrote `postage_tax_strategy`. The screen landed in thelia-templates/default-twig#37: a "Postage tax breakdown" section on the tax rules page, the three strategies as named and explained radio options, a value the enum does not know refused rather than silently degraded, and the change recorded in the admin log.

This adds the back-office end-to-end coverage: the three options render with `single_rule` checked by default, and a saved strategy comes back checked after the round trip, the run putting the shop back on its original behaviour. Verified against a running install with `BO_TEMPLATE=default-twig`, 9 passed on this page; the save test was checked red by removing the `ConfigQuery::write`.

### DIFF
--- a/tests/Playwright/poms/backoffice/taxes-rules-page.ts
+++ b/tests/Playwright/poms/backoffice/taxes-rules-page.ts
@@ -65,4 +65,20 @@ export class TaxesRulesPage extends BaseAdminPage {
   postageTaxRuleRow(moduleCode: string): Locator {
     return this.page.getByTestId(`postage-tax-rule-row-${moduleCode}`);
   }
+
+  get postageTaxStrategySection(): Locator {
+    return this.page.getByTestId('postage-tax-strategy-section');
+  }
+
+  get postageTaxStrategyForm(): Locator {
+    return this.page.getByTestId('postage-tax-strategy-form');
+  }
+
+  get postageTaxStrategySave(): Locator {
+    return this.page.getByTestId('postage-tax-strategy-save');
+  }
+
+  postageTaxStrategyOption(strategy: string): Locator {
+    return this.page.locator(`#postage-tax-strategy-${strategy}`);
+  }
 }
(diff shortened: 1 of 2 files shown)


anoziere:
The `test` job fails on `main` too, and on the same two errors only:

```
```

`main` at 9e29412c — https://github.com/thelia/thelia/actions/runs/31690677335 — produces that list and nothing else, as does this branch. The Propel base classes of the table #3721 added are generated from `thelia/config`, which has not been released yet; `composer ci` stops at PHPStan, so the PHPUnit suites do not run on either side.

This branch only touches `tests/Playwright`, which PHPStan does not analyse: cs-diff passes ahead of PHPStan, and `phpstan-botwig` is green.

PR #3733, opened 2026-08-13
fix: make the front address endpoints usable and safe

### LINKED ISSUE #3728: Creating an address through the API fails when address2 or address3 is omitted
`address2` and `address3` are declared as non-nullable strings with no default:

- `core/lib/Thelia/Api/Resource/Address.php:158` — `public string $address2;`
- `core/lib/Thelia/Api/Resource/Address.php:161` — `public string $address3;`

A payload that omits them leaves the properties uninitialized, and the transformer skips uninitialized properties (`core/lib/Thelia/Api/Bridge/Propel/Service/ApiResourcePropelTransformerService.php:344`), so the columns never reach the insert. Both are `NOT NULL` without a default in the schema, so MySQL answers `SQLSTATE[HY000]: General error: 1364 Field 'address2' doesn't have a default value` and the caller gets a `500`.

Reproduced on `POST /api/admin/addresses` and `POST /api/front/account/addresses` with a complete payload minus `address2`/`address3`.

The two lines are optional on every address form, so a client should not have to send an empty string for them. A default value on the resource properties is enough — no schema change.

### LINKED ISSUE #3729: The front address endpoints take the owner from the request body
`/front/account/addresses` takes the owner of the address from the request body:

- `core/lib/Thelia/Api/Resource/Address.php:198-200` — `customer` carries `self::GROUP_FRONT_WRITE`, so a front caller sets it;
- `core/lib/Thelia/Api/Resource/Address.php:67-70` and `:78-81` — the front `Post` and `Put` use the generic persist processor, which never looks at the token.

Two consequences, both reproduced against a database seeded by `bin/test-prepare`:

1. **An authenticated customer writes into another customer's address book.** `POST /api/front/account/addresses` with `"customer": "/api/admin/customers/<victim id>"` answers `201`, and the row lands on the victim: `AddressQuery::create()->filterByCustomerId($victim->getId())->count()` returns 1. The victim then sees an address they never created, and the read side is scoped to the owner, so they cannot tell where it came from.
2. **A caller that sends no `customer` gets a `500`.** `address.customer_id` is `NOT NULL`, so the insert fails with `SQLSTATE[HY000]: General error: 1364`. Nothing is written — no orphan row — but the endpoint is unusable as documented.

The `Get`, `Put` and `Delete` item operations already carry `security: 'object.customer.getId() == user.getId()'`; only the creation path was left open. An account endpoint should take its owner from the token and ignore what the body claims.

### LINKED ISSUE #3731: Front address writes skip every validation constraint
Every constraint on the address resource is declared for the admin write groups only, so a front write is validated against nothing:

- `core/lib/Thelia/Api/Resource/Address.php:439` — `#[Callback(groups: [self::GROUP_ADMIN_WRITE, Customer::GROUP_ADMIN_WRITE])] verifyZipcode`
- `core/lib/Thelia/Api/Resource/Address.php:142`, `:146`, `:150`, `:154`, `:165`, `:178`, `:192` — the `NotBlank` constraints carry the same two groups.

The front operations denormalize with `self::GROUP_FRONT_WRITE` (`:86`), and the transformer validates with the denormalization groups (`core/lib/Thelia/Api/Bridge/Propel/Service/ApiResourcePropelTransformerService.php:153`), so none of them run.

Measured on a seeded database:

- `POST /api/front/account/addresses` with `"zipcode": "75"` on France, seeded with the `NNNNN` format, answers `201`. The same payload on `/api/admin/addresses` answers `422`.
- Omitting `label` answers `500` (the column is `NOT NULL`) where the admin endpoint answers `422`.

A front client should get the same answers as the back office, and a missing required field is a validation error, not a database crash.

anoziere:
Three defects on the same write path, each proved failing first.

### The owner came from the body — #3729

`POST /api/front/account/addresses` with another customer's IRI answered `201` and filed the address in that customer's address book; without a `customer` it reached the database with no owner and crashed on a `NOT NULL` column. `CustomerAddressProcessor` now reads the owner from the token on the front `Post` and `Put`, and `customer` leaves the front write group so the body cannot name it. The item operations already carried `security: 'object.customer.getId() == user.getId()'`; only the creation path was open.

### Front writes were validated against nothing — #3731

Every constraint carried the admin groups only, while the front operations denormalize with `front:address:write` and the transformer validates with the denormalization groups. A zip code of `75` on France was stored as sent, and a missing `label` answered `500` where the back office answers `422`. The `NotBlank` constraints and the zip-code callback now carry the front write group too, and `customerTitle` gains the `NotBlank` it never had.

### The optional address lines could not be omitted — #3728

`address2` and `address3` are non-nullable strings with no default, so omitting them left the properties uninitialized, the transformer skipped them, and MySQL rejected the insert. They now default to an empty string — no schema change.

### Tests

`tests/Api/Front/AddressApiTest.php` and `tests/Api/Admin/AddressApiTest.php`: the created address belongs to the caller, another customer's address book stays empty, the zip code is checked against the country format, a missing required field is a `422`, and the optional lines may be omitted on both sides. All seven were checked failing against `main` before the fix.

### Behaviour change

A front client that relied on sending `customer` will find the field ignored, and payloads the endpoint used to accept without checking now answer `422`.

Fixes #3728
Fixes #3729
Fixes #3731

### DIFF
--- a/core/lib/Thelia/Api/Resource/Address.php
+++ b/core/lib/Thelia/Api/Resource/Address.php
@@ -32,6 +32,7 @@ use Thelia\Api\Bridge\Propel\Attribute\Relation;
 use Thelia\Api\Bridge\Propel\Filter\BooleanFilter;
 use Thelia\Api\Bridge\Propel\Filter\NotInFilter;
 use Thelia\Api\Bridge\Propel\Filter\SearchFilter;
+use Thelia\Api\State\Processor\CustomerAddressProcessor;
 use Thelia\Core\Translation\Translator;
 use Thelia\Model\Map\AddressTableMap;
 
@@ -66,6 +67,7 @@ use Thelia\Model\Map\AddressTableMap;
         new Post(
             uriTemplate: '/front/account/addresses',
             normalizationContext: ['groups' => [self::GROUP_FRONT_READ, self::GROUP_FRONT_READ_SINGLE]],
+            processor: CustomerAddressProcessor::class,
         ),
         new GetCollection(
             uriTemplate: '/front/account/addresses',
@@ -78,6 +80,7 @@ use Thelia\Model\Map\AddressTableMap;
         new Put(
             uriTemplate: '/front/account/addresses/{id}',
             security: 'object.customer.getId() == user.getId()',
+            processor: CustomerAddressProcessor::class,
         ),
         new Delete(
             uriTemplate: '/front/account/addresses/{id}',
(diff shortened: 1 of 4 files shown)


anoziere:

```
```

Same list, in the same order, on main: https://github.com/thelia/thelia/actions/runs/31691613513. `local/config/schema.xml` has neither table yet, so the Base classes cannot be generated in CI; it clears with the `thelia/config` release at the end of the wave. Neither model belongs to this branch — they arrive through the merge commit CI builds. PHPStan runs before the suites in the `ci` script, so the five PHPUnit suites did not start here either.

Run locally on this branch against a database prepared by `bin/test-prepare`: `cs-diff` 0 of 1849, `phpstan` 0 errors, unit 284 green, api 213 green (1 skip), integration 567 of 571 — the four remaining are `GaplessSequenceGeneratorTest`, which opens its own connection to the development database in that sandbox.

Issue #3732, opened 2026-08-13
Front collections advertise admin IRIs for their members


anoziere:
A front collection answers with members whose `@id` points at the admin surface:

```
GET /api/front/languages
"hydra:member":[{"@id":"/api/admin/languages/1","@type":"Lang",...}]
```

Same on `GET /api/front/customer_titles`, whose members carry `/api/admin/customer_titles/1`.

Resources declare their front and admin sides in two `#[ApiResource]` blocks on the same class, and the IRI converter resolves one item operation per class, so the admin one wins whatever the operation being served. A headless client following those links reaches an endpoint it is not allowed to call, and the response tells it about a surface it should not see.

Found while adding the front item operation for customer titles (#3730), which does not change this: the item is reachable at `/api/front/customer_titles/{id}`, but the collection still advertises the admin URL.

Not investigated further: the fix belongs to the IRI generation, not to a single resource.

PR #3734, opened 2026-08-13
fix: read a single customer title from the front API

### LINKED ISSUE #3730: No front API operation reads a single customer title
The front side of the customer title resource declares a collection and nothing else:

- `core/lib/Thelia/Api/Resource/CustomerTitle.php:57-64` — one `GetCollection` on `/front/customer_titles`.

An address, like a registration, has to name a title, and a client can only do that with an IRI the front API answers to. `GET /api/front/customer_titles/1` returns `404`, so a front payload has to borrow the admin IRI `/api/admin/customer_titles/1` — a URL under a surface the client has no business knowing, and one it cannot read.

Adding the item operation is enough; the read group already exists (`self::GROUP_FRONT_READ`).

Related, and not fixed here: the front collection hands out admin IRIs in its members (`"@id": "/api/admin/customer_titles/1"` on `GET /api/front/customer_titles`). That one is not specific to customer titles — see #3732.

anoziere:
`GET /api/front/customer_titles/1` answered `404`: the front side of the resource declared a collection and nothing else, so the only readable item was `/api/admin/customer_titles/1`. A front client naming the title of an address or of a registration had to quote a URL under a surface it cannot call.

The item operation reuses the existing `front:customer_title:read` group, so the payload is the one the collection already returns.

`tests/Api/Front/CustomerTitleApiTest.php` was checked failing against `main` first.

Not fixed here: the collection still advertises admin IRIs in its members, which is not specific to customer titles — #3732.

Fixes #3730

### DIFF
--- a/core/lib/Thelia/Api/Resource/CustomerTitle.php
+++ b/core/lib/Thelia/Api/Resource/CustomerTitle.php
@@ -59,6 +59,12 @@ use Thelia\Model\Map\CustomerTitleTableMap;
         new GetCollection(
             uriTemplate: '/front/customer_titles',
         ),
+        // Without a front item operation, the collection hands out admin IRIs
+        // and a front client has no title to name on an address or a
+        // registration.
+        new Get(
+            uriTemplate: '/front/customer_titles/{id}',
+        ),
     ],
     normalizationContext: ['groups' => [self::GROUP_FRONT_READ]],
 )]
--- a/tests/Api/Front/CustomerTitleApiTest.php
+++ b/tests/Api/Front/CustomerTitleApiTest.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+/*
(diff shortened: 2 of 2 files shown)


anoziere:

```
```

Same list, in the same order, on main: https://github.com/thelia/thelia/actions/runs/31691613513. `local/config/schema.xml` has neither table yet, so the Base classes cannot be generated in CI; it clears with the `thelia/config` release at the end of the wave. Neither model belongs to this branch — they arrive through the merge commit CI builds. PHPStan runs before the suites in the `ci` script, so the five PHPUnit suites did not start here either.

Run locally on this branch against a database prepared by `bin/test-prepare`: `cs-diff` 0 of 1849, `phpstan` 0 errors, unit 284 green, api 213 green (1 skip), integration 567 of 571 — the four remaining are `GaplessSequenceGeneratorTest`, which opens its own connection to the development database in that sandbox.

Issue #3707, opened 2026-08-13
A service declared public in Config/Resources/services can end up private in the container


anoziere:
A service declared `public()` in `core/lib/Thelia/Config/Resources/services/**.php` can end up private in the compiled container. Nothing warns about it: the declaration is there, it just does not survive the container build, and the failure only shows up at runtime the first time something calls `$container->get()` on that service.

## What happens

`TheliaKernel` loads `Config/Resources/services.php` twice:

- `TheliaKernel::configureContainer()` imports `Config/Resources/*.php` — that is `services.php` — then every `Config/Resources/services/**/*.php` file (`core/lib/Thelia/Core/TheliaKernel.php:188-197`);
- `TheliaKernel::buildContainer()` calls `loadService()`, which loads `services.php` a second time (`core/lib/Thelia/Core/TheliaKernel.php:366` and `:511-516`).

`services.php` registers the whole `Thelia\` namespace by PSR-4 (`core/lib/Thelia/Config/Resources/services.php:55-63`). On the second pass that registration runs again, after the `services/**` files, and replaces their definitions with fresh autowired, **private** ones. Only the classes those files configure by FQCN are affected; aliases and string service ids are untouched, which is why most of the container looks fine.

## Proof

`Thelia\Core\Serializer\SerializerManager` is declared public in `core/lib/Thelia/Config/Resources/services/core/serializers.php:25-29`, and `Thelia\Core\Archiver\ArchiverManager` in `services/handlers/archiver.php:26-30`. The compiled container disagrees:

```
$ php bin/console debug:container Thelia\\Core\\Serializer\\SerializerManager
  Service ID       Thelia\Core\Serializer\SerializerManager
  Public           no
  Autowired        yes          <- the signature of the PSR-4 load, not of the set() in serializers.php
```

The public *aliases* (`thelia.serializer.manager`, `thelia.archiver.manager`) survive, since the load never touches those ids. Hence the misleading error message users get:

```
The "Thelia\Core\Serializer\SerializerManager" service or alias has been removed or
inlined when the container was compiled. You should either make it public, or stop
using the container directly and use dependency injection instead.
```

That is exactly what broke the export console command (#3675): every `export` invocation except `--list-export` died on it, on every reference. The workaround already present in the codebase is `services.php:142-143`, where `ConfigCacheService` is marked public *after* the load — the one service that gets its visibility back.

## Scope

Replaying the two loading passes over the current `main` config gives three definitions that lose the flag they were given:

[...]

PR #3735, opened 2026-08-13
fix: make generate:sql runnable again and the seed reproducible

### LINKED ISSUE #3725: generate:sql cannot run on Thelia 3, so insert.sql.tpl has drifted from insert.sql

anoziere:
Fixes #3725

`generate:sql` failed on every Thelia 3 install: it asked the container for a private `thelia.parser`, and even public that service could not have helped — the command drove `unregisterPlugin()` / `registerPlugin()` on a Smarty parser core no longer ships. So nothing could regenerate `setup/insert.sql`, and the seed drifted from `setup/insert.sql.tpl` for the whole 3.0 line with no way to notice.

## The generator

`setup/insert.sql.tpl` becomes a Twig template — Twig is what Thelia 3 renders with, and what `Message::buildMessage()` already uses for the wording this file seeds. The grammar it needed was small: a loop over the locales, a `loop.last` comma, and 1609 `intl()` calls.

Rendering moves to `Thelia\Install\Generator\SeedSqlGenerator`, which the command injects. No service has to be made public: the generator and the `Translator` it needs are ordinary autowired services, so the `$container->get()` that #3707 would have bitten is gone rather than patched. The generator builds its own Twig environment with autoescaping off — the seed is SQL, and the connection quotes it — instead of adding an `intl()` function to the application's.

`generate:sql` now defaults to the locales the seed creates, read from the `lang` block of the template, so a plain run reproduces the shipped file. Seeding the wording of a language the shop does not have would only bloat the file, and since #3697 a language added later gets its wording replayed anyway. `--locales` still overrides, and `--check` reports a difference without writing.

## The template

The drift was not only the placeholders of #3724. Regenerating showed the template still describing Thelia 2.6, so it is brought back in line with the shipped seed — which is the reference:

- `active-front-template` is `flexy`, not `default`
- no Dutch language, no `module_i18n` block, no `area_delivery_module` rows
- `Thelia\Domain\Taxation\TaxEngine\TaxType\PricePercentTaxType`
- the current `not_paid` and `canceled` colours
- the API hooks (1316-1318) and the API resource (48) are gone
- `item.edition.images` (1425) and the `customer_send_code` message are added, with their wording added to `setup/I18n` so the seed can produce it

## What regenerating restores

The seed had been hand-maintained, and hand-maintaining lost wording the catalogs had all along. Regenerating puts it back: 249 Italian country names where the file had `NULL`, Spanish, Italian and Russian state names, the Czech row for Taiwan, and `Côte d'Ivoire` whose apostrophe had been eaten into `Côte dIvoire`. The `thelia_version` rows go from `2.6.0` to the running version — `DatabaseSetup::writeTheliaVersion()` overwrites them at install precisely because they had gone stale.

Applied to an empty database on top of `setup/thelia.sql`, the regenerated seed loads clean.

## Tests

`tests/Integration/Install/SeedSqlGeneratorTest.php` is the acceptance test the issue asked for: it regenerates the seed and asserts it is byte for byte the file on disk, and that the seeded locales are the languages the seed creates. Changing one Italian country name in `setup/I18n` turns it red with the diff and the command to run:

```
setup/insert.sql is not what setup/insert.sql.tpl and setup/I18n produce. Run: php Thelia generate:sql
-    (1, 'it_IT', 'Afghanistan', NULL, NULL, NULL),
+    (1, 'it_IT', 'Afganistan', NULL, NULL, NULL),
```

`SeedPlaceholderSyntaxTest` keeps the two catalog guards and follows the template to its Twig syntax; its mail-subject comparison goes, since the new test proves the same thing for every table rather than for `message_i18n` alone. 8 unit and 8 integration seed tests green, `php-cs-fixer` 0 of 1848, PHPStan clean.

## Not in scope

`setup/update/tpl/*.tpl` stays as it is and the command no longer renders it. Those 65 templates resolve their wording against today's `setup/I18n`, so regenerating a 2.0 update script would rewrite a shipped artifact with a decade of translation edits. Update scripts have been written straight as `.sql` since 2.5.0-alpha2 — ten files with no template — so the directory is history, not a source. Worth deleting in its own change.


anoziere:
Merging with `--admin`: the red `test` check is the `thelia/config` staleness of this wave, not this branch.

The whole failure is PHPStan missing generated Propel Base classes for two tables merged earlier in the wave:

```
```

`order_postage_tax` comes from #3721 and `product_sale_elements_virtual_document` from #3520; both wait on the config release that closes the wave. The same run is red on `main` and on every open PR, and the failure is identical here to #3726.

This branch touches `setup/`, `tests/`, one command and one new generator class. `cs-diff` passes in the same run (0 of 1854), and PHPStan reports these five errors and nothing else. PHPStan gates before PHPUnit, so no suite ran; locally the 8 unit and 8 integration seed tests are green, the regenerated seed loads clean into an empty database on top of `setup/thelia.sql`, and changing one Italian country name in `setup/I18n` turns the new acceptance test red with the diff and the command to run.

PR #3737, opened 2026-08-13
chore: drop the update script templates nothing renders

### LINKED ISSUE #3736: setup/update/tpl holds 65 templates nothing renders and nothing should
`setup/update/tpl` holds 65 Smarty templates for the update scripts of Thelia 2.0 to 3.0.0-beta1. Nothing renders them any more, and nothing should.

`generate:sql` stopped writing them in #3735, and it could not have kept doing so: they are Smarty, and core no longer ships a parser for it. Update scripts have been written straight as `.sql` since 2.5.0-alpha2 — ten files in `setup/update/sql` have no template at all.

Regenerating one would be worse than useless. Their `{intl}` calls resolve against today's `setup/I18n`, which has taken ten years of edits, so a 2.0 update script generated now would not be the script that shipped in 2014. The output would silently rewrite an artifact that has already run on every shop that went through that version.

Nothing reads the directory: `Update::SQL_DIR` is `update/sql/`, and no code, config, packaging or CI file refers to it.

Leaving 65 templates that look like sources but reproduce nothing is the state that let `insert.sql.tpl` drift for a whole major version without anyone noticing (#3725). The directory is history and belongs in the history.

anoziere:
Fixes #3736

Removes the 65 Smarty templates of `setup/update/tpl`, which produced the update scripts of Thelia 2.0 to 3.0.0-beta1.

`generate:sql` stopped rendering them in #3735, and it could not have kept doing so: they are Smarty and core no longer ships a parser for it. Update scripts have been written straight as `.sql` since 2.5.0-alpha2 — ten files in `setup/update/sql` have no template at all.

Regenerating one would have been actively harmful. Their `{intl}` calls resolve against today's `setup/I18n`, so a 2.0 script generated now would not be the script that shipped, and the output would silently rewrite an artifact that has already run on every shop that went through that version. `setup/update/sql` is untouched: the scripts themselves stay.

## The sweep

Nothing reads the directory.

- `Thelia\Core\Install\Update::SQL_DIR` is `update/sql/`, and `setup/update.php` drives that class — neither has any notion of `tpl`.
- `grep -rn "update/tpl"` over the repository returns one hit, a 2016 CHANGELOG line about a fix made in one of these files.
- Every other `.tpl` reference in PHP, XML, YAML, JSON, NEON and shell points at `setup/insert.sql.tpl`, which stays and remains the source of the seed, except `TlogDestinationPopup.tpl` (unrelated) and a `->notName('*.tpl')` exclusion in `.php-cs-fixer.dist.php` that stays valid for `insert.sql.tpl`.
- `setup/composer.json` ships the package with no file list, `.gitattributes` and the CI workflows never mention `tpl`.

`composer test`, `cs-diff` and PHPStan are unaffected: no code changes, only deletions of files nothing loads.


anoziere:
Merging with `--admin`: the red `test` check is the `thelia/config` staleness of this wave, not this branch.

The whole failure is PHPStan missing generated Propel Base classes for two tables merged earlier in the wave:

```
```

`order_postage_tax` comes from #3721 and `product_sale_elements_virtual_document` from #3520; both wait on the config release that closes the wave. The same run is red on `main` and on every open PR, identical to #3726 and #3735.

This branch adds and changes no code at all — it deletes 65 template files that nothing loads. `cs-diff` passes in the same run (0 of 1854), and PHPStan reports these five errors and nothing else.

PR #3738, opened 2026-08-13
fix: name an item on the surface the request was served from

### LINKED ISSUE #3732: Front collections advertise admin IRIs for their members

anoziere:
Fixes #3732

### The cause

`ApiPlatform\Symfony\Routing\IriConverter::getIriFromResource()` re-resolves an operation whenever the one it is given cannot name an item — none at all, a collection, or the `POST` that just created the resource:

- `vendor/api-platform/symfony/Routing/IriConverter.php:149-151` — with no operation it builds a bare `Get`, which has no name;
- `:160-170` — a nameless operation is then resolved through `resourceMetadataCollectionFactory->create($class)->getOperation(null, false, true)`, the **first item operation of the class**.

Resources declare their admin and their front side in two `#[ApiResource]` blocks on the same class, admin first, so that lookup always lands on the admin operation. Instrumented on `GET /api/front/languages`: the collection `@id` is generated with the front `GetCollection` and is correct, then every member arrives with `operation = null` and no `item_uri_template`, and is named `/api/admin/languages/1`.

The same path names relations, so a front item leaked admin IRIs for everything it embedded: `GET /api/front/products/{id}` returned `"taxRule": {"@id": "/api/admin/tax_rules/1"}`.

### The fix

`SurfaceAwareIriConverter` decorates `api_platform.iri_converter`. When the converter is about to resolve an operation on its own for an item, it looks one up under the first path segment of the operation being served (`_api_operation` on the current request) and hands it over; everything else — a class asked for its collection IRI, an explicit `item_uri_template`, a request outside API Platform — passes straight through. Nothing resource-specific, and no change to the admin surface.

`Lang` was the only resource with a front collection and no front item operation, so its members had nothing else to be named by. It gains `GET /front/languages/{id}` on `front:lang:read`, the group its collection already returns, so no field becomes readable that was not already.

### Consequences, swept before merging

- **Flexy** reads through `resources()` server-side, where no `_api_operation` is set, so the decorator is a no-op there; the theme contains no `/api/admin/` string and never reads `@id` (checked, not modified — the theme is frozen).
- The **Twig back office** calls no `/api/front/` endpoint, and admin answers are unchanged.
- No module and no test parses an `@id` out of a front response; `I18nDenormalizer` resolves IRIs in the read direction, which this decorator passes through untouched.
- A client that stored an admin IRI received from a front answer was storing a URL it could not call in the first place.

### Tests

`tests/Api/Contract/SurfaceIriContractTest.php`, checked failing first — the three failures were `/api/admin/languages/1`, `/api/admin/customer_titles/1` and the embedded `/api/admin/tax_rules/1`. It covers a front collection, the relations of a front item, and an admin collection that must keep its admin IRIs.

### DIFF
--- a/core/lib/Thelia/Api/Bridge/Propel/Routing/SurfaceAwareIriConverter.php
+++ b/core/lib/Thelia/Api/Bridge/Propel/Routing/SurfaceAwareIriConverter.php
@@ -1,0 +1,144 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Thelia package.
+ * http://www.thelia.net
+ *
+ * (c) OpenStudio <info@thelia.net>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Thelia\Api\Bridge\Propel\Routing;
+
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\HttpOperation;
+use ApiPlatform\Metadata\IriConverterInterface;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\Metadata\UrlGeneratorInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Keeps an IRI on the surface the request is being served from.
+ *
+ * A resource declares its admin and its front side in two #[ApiResource] blocks
+ * on the same class. When the serializer asks for the IRI of an item it has no
+ * operation to hand — a member of a collection, a relation, the body of a POST —
+ * and API Platform then resolves the first item operation of the class, which is
+ * the admin one. A front collection therefore answered with admin IRIs: links a
+ * front client cannot follow, pointing at a surface it should not know about.
+ *
+ * The operation being served says which surface the answer belongs to, so the
+ * item operation is looked up under the same first path segment.
+ */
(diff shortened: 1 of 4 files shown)


anoziere:

```
```

Same list on main: https://github.com/thelia/thelia/actions/runs/31693512358. `local/config/schema.xml` has neither table yet, so the Base classes cannot be generated in CI; it clears with the `thelia/config` release at the end of the wave. Neither model belongs to this branch — they arrive through the merge commit CI builds. PHPStan runs before the suites in the `ci` script, so the five PHPUnit suites did not start here either.

Run locally on this branch against a database prepared by `bin/test-prepare`: `cs-diff` 0 of 1852, `phpstan` 0 errors, unit 288 green, api 217 green (1 skip, plus a warning and a deprecation from `maiorano84/shortcodes` that predate this branch), integration 576 of 580 — the four remaining are `GaplessSequenceGeneratorTest`, which opens its own connection to the development database in that sandbox.

Issue #3774, opened 2026-08-18
A successful update is reported as an error, and the operator is offered a restore that destroys the migration


anoziere:
## What happens

Upgrading a 3.0.0-beta2 shop to beta3 applies everything correctly — the SQL script, the PHP script, and the `thelia_version` marker — and then reports the run as a failure:

```
Sorry, an unexpected error has occured : There is no active transaction
#0 setup/update.php(161): Thelia\Core\Install\Update->process()
```

`Update::process()` opens a transaction before the loop and commits it after. Any `CREATE`, `ALTER` or `DROP` in an update script makes MariaDB commit implicitly and end that transaction, so the `commit()` finds nothing to commit and raises.

## Reproduction

On an installed shop:

```sql
UPDATE config SET value='3.0.0-beta2' WHERE name='thelia_version';
UPDATE config SET value='beta2'       WHERE name='thelia_extra_version';
```

then `php setup/update.php`, answering yes to the update and yes to the backup.

## Why the transaction cannot hold

Run against the connection the update loop uses:

```php
$connection->beginTransaction();
$connection->exec('INSERT INTO probe_a (id) VALUES (1)');           // DML, inside the transaction
$connection->exec('CREATE TABLE probe_b (id INT) ENGINE=InnoDB');   // DDL, implicit commit
$connection->exec('INSERT INTO probe_a (id) VALUES (2)');           // after the implicit commit
$connection->rollBack();
```

```
after beginTransaction  : inTransaction=true
after DDL               : inTransaction=false
rollBack()              : PDOException: There is no active transaction
rows surviving rollBack : 2      <- both, including the one written before the DDL
commit()                : PDOException: There is no active transaction
```

The row written *before* the DDL survives the rollback. So once a script touches the schema, the transaction protects nothing at all — not even the statements that preceded the DDL in the same file. 33 of the 76 shipped `setup/update/sql/*.sql` files contain DDL, and `3.0.0-beta2.sql` and `3.0.0-beta3.sql` add more through guarded `PREPARE`/`EXECUTE`, so every upgrade path that reaches 3.0 crosses it.

The rest of `process()` already knows this: both `rollBack()` calls are guarded by `inTransaction()`. Only the `commit()` is not.

## Two consequences

**1. Every successful schema migration is reported as an error.** The shop is fully migrated, `thelia_version` is correct, and the operator is told the update failed. `setup/update.php` exits 7.

[...]

PR #3778, opened 2026-08-18
fix: make the backup taken before an update restorable

### LINKED ISSUE #3777: The backup taken before an update cannot be restored: NULLs are dumped as empty strings
## What happens

The backup `setup/update.php` takes before an update cannot be restored on a shop with realistic data. Accepting the restore stops partway:

```
Database restore started. Wait, it could take a while...
SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect integer value: '' for column `address`.`state_id` at row 1
Sorry, your database can't be restore. Try to do it manually : local/backup/update.sql
```

The dump is a sequence of `DROP TABLE` + `CREATE TABLE` + `INSERT` per table, applied in `SHOW TABLES` order, so it fails **after** having already dropped and rebuilt the tables it got through. On a demo install it dies on `address`, which is left empty, while later tables keep their pre-restore content. The database ends up in a state that is neither the backup nor what it was before.

This makes the backup useless as the safety net it is offered as — it is the only mechanism that can undo a schema migration, and it is the thing the operator is pointed at when an update fails.

## Cause

`Database::backupDb()` serialises every value the same way:

```php
$row[$j] = addslashes((string) $row[$j]);
$row[$j] = str_replace("\n", '\\n', $row[$j]);
$data[] = '"'.$row[$j].'"';
```

`(string) null` is `''`, so a NULL is written as `""`. Re-inserting `""` into a nullable `INTEGER` column is rejected by MariaDB under the default `STRICT_TRANS_TABLES`. `address.state_id` is nullable and null for most rows, hence the first failure; every nullable non-string column in the schema has the same problem.

Two more consequences of the same loop, on the same values:

- `addslashes()` escapes `'`, `"`, `\` and NUL but is not the connection's quoting function, so it does not account for the connection charset; `quote()` on the connection, or a prepared insert, is what matches the data going back in.
- `str_replace("\n", '\n', …)` turns a real newline into the two characters `\` and `n`, which then read back as a newline only because the value sits in a double-quoted SQL string — but a literal backslash already in the data has been doubled by `addslashes()` first, so `\n` typed by a user comes back as a newline.

## Reproduction

On a demo install, set the version marker back one release, run `php setup/update.php`, accept the backup, and accept the restore when the run reports a failure. Or, without an update at all, call `Database::backupDb()` on any table holding a NULL in a non-string column and replay the file.

Checking the produced dump is enough to see it:

```
$ grep -c NULL local/backup/update.sql
1182            # the string NULL, inside data, never as a value
$ grep -o 'INSERT INTO `address` VALUES([^)]*)' local/backup/update.sql | head -1
INSERT INTO `address` VALUES("1","Main address","1","1","","thelia", …
                                                        ^^ state_id, NULL in the source
```

## Note

The false failure that used to send operators into this restore path in the first place is #3774. Fixing that keeps the restore away from successful migrations, but the restore still has to work when a migration really does fail, which is when it matters.

anoziere:
Fixes #3777.

The update loop no longer wraps itself in a transaction MariaDB cannot honour (#3774), which leaves the backup taken before the run as the only way back. It could not be restored. Two independent causes, both silent.

### NULL was dumped as an empty string

Every value went through `addslashes((string) $value)` and was written between double quotes, so a NULL became `""`. Under `STRICT_TRANS_TABLES` that is rejected for any non-string column, and the restore died on the first one it met — having already dropped and rebuilt the tables before it, and emptied the one it was on:

```
SQLSTATE[22007]: Incorrect integer value: '' for column `address`.`state_id` at row 1
```

A full dump and restore of a demo shop lost the whole `address` table, 12 rows, and stopped there.

Values are now quoted by the connection, which knows the charset it talks in and the escaping mode of the server, and NULL is written as `NULL`. That also settles the newlines: the connection escapes CR, LF and CTRL-Z itself, so the hand-rolled `str_replace("\n", '\n', …)` that used to follow `addslashes()` is gone rather than repaired — nothing can reach the file as a raw newline and split a statement. For the record, that substitution was *not* broken: `addslashes()` ran first, so a backslash already in the data was doubled before the newline escape was inserted, and both round-tripped. The claim to the contrary in the issue was wrong; the tests below pin the behaviour either way.

### Every pipe in the data came back a semicolon

`prepareSql()` hides the semicolons inside a `DELIMITER` block behind a placeholder while it cuts the file into statements on `";\n"`, then puts them back. The placeholder was `|`, and the reverse substitution ran over **every** statement, not only the ones that came from such a block. So any pipe a shop had written was turned into a semicolon on the way in.

The demo data alone hits it — the rich text toolbar in `config` is a pipe-separated list, and it came back with semicolons. This affects `insertSql()` in general, so it also applies to any seed file whose data holds a pipe.

The placeholders are now drawn per call and checked against the file, so no value can be mistaken for one on the way back. The same treatment covers the older `-CODE-` marker, which had the same flaw for text containing `-CODE-`. `thelia.sql` ships `DELIMITER` blocks, and a fresh install still creates its routine — there is a test for that too.

### A restore that stops no longer stops in silence

A dump is a `DROP`, a `CREATE` and the rows of one table after another, and DDL cannot be rolled back, so a restore that fails halfway leaves a database that is neither the backup nor what it was. True atomicity is out of reach here — it would mean restoring into a scratch schema and swapping it in with a single `RENAME TABLE`, which is a different design and not this fix. What is in reach, and is what this does:

- **the dump is checked before anything is dropped.** A file that does not end the way a complete dump ends is refused outright, so a backup cut short by a full disk or a killed process is never replayed. Nothing is touched.
- **a failure says where it stopped:** `The restore stopped while rebuilding \`address\`. The tables restored before it now hold the backup, the ones after it still hold what they held: …`, surfaced through `Update::getRestoreFailure()` and printed by `update.php`, which used to swallow it in an `echo` from inside the domain class.
- **foreign key checks are put back** whatever happens. A dump turns them off to be replayable in any order and on again at its end; a restore that stopped in between used to leave them off on the connection.

### Verified

A dump and restore of a demo install — 162 tables, 15,074 rows — now comes back identical, compared row by row rather than by `CHECKSUM TABLE`, which is not a content-only function and disagreed while every cell matched. On `main` the same run dies on `address`.

End to end: an update that fails now offers the backup, and accepting it brings the shop back — 12 addresses, 10 customers, 30 products, the marker on its previous version, and the pipes intact.

`tests/Integration/Install/DatabaseBackupTest.php` round-trips the values that break naive escaping, one per case: NULL in a nullable integer column, literal backslashes, a `\n` that is not a newline, real newlines and CRLF, both kinds of quote, a semicolon before a newline, multibyte UTF-8, four-byte emoji, SQL that would run if its quotes escaped, control characters, the empty string, pipes, and text shaped like the internal markers. Seven of the eighteen tests fail on `main`.

### DIFF
--- a/core/lib/Thelia/Core/Install/Database.php
+++ b/core/lib/Thelia/Core/Install/Database.php
@@ -27,6 +27,9 @@ use Thelia\Log\Tlog;
  */
 class Database
 {
+    /** The statement backupDb() closes a dump with, and restoreDb() looks for to tell a whole file from a truncated one. */
+    private const DUMP_TERMINATOR = 'SET foreign_key_checks=1;';
+
     protected ConnectionInterface|\PDO $connection;
 
     /**
@@ -122,7 +125,15 @@ class Database
      */
     protected function prepareSql($sql): array
     {
-        $sql = str_replace(";',", '-CODE-', $sql);
+        // Both substitutions below hide a semicolon from the splitter and put it back
+        // once the file is cut. The tokens are drawn per call and checked against the
+        // file, so nothing in the data can be mistaken for one on the way back. Fixed
+        // markers could be: the placeholder was '|', so every pipe a shop had written
+        // came out of a dump as a semicolon.
+        $quotedSemicolon = $this->placeholderAbsentFrom($sql);
+        $blockSemicolon = $this->placeholderAbsentFrom($sql);
+
+        $sql = str_replace(";',", $quotedSemicolon, $sql);
         $sql = trim($sql);
         preg_match_all('#DELIMITER (.+?)\n(.+?)DELIMITER ;#s', $sql, $m);
 
@@ -131,20 +142,30 @@ class Database
                 throw new \RuntimeException('You can not use "|" as delimiter: '.$v);
             }
 
-            $stored = str_replace([';', $m[1][$k]], ['|', ";\n"], $m[2][$k]);
+            $stored = str_replace([';', $m[1][$k]], [$blockSemicolon, ";\n"], $m[2][$k]);
             $sql = str_replace($v, $stored, $sql);
         }
 
         $query = [];
 
-        $tab = explode(";\n", $sql);
-
-        foreach ($tab as $iValue) {
-            $queryTemp = str_replace(['-CODE-', '|'], [";',", ';'], $iValue);
-            $query[] = $queryTemp;
+        foreach (explode(";\n", $sql) as $iValue) {
+            $query[] = str_replace([$quotedSemicolon, $blockSemicolon], [";',", ';'], $iValue);
         }
 
         return $query;
+    }
+
+    /**
+     * A token the file does not already contain, so that putting it back cannot alter
+     * a value that happened to look like it.
+     */
+    private function placeholderAbsentFrom(string $sql): string
+    {
+        do {
+            $placeholder = '{{thelia-sql-'.bin2hex(random_bytes(8)).'}}';
+        } while (str_contains($sql, $placeholder));
+
+        return $placeholder;
     }
 
     /**
@@ -202,9 +223,7 @@ class Database
                     $data[] = 'INSERT INTO `'.$table.'` VALUES(';
 
                     for ($j = 0; $j < $fieldCount; ++$j) {
-                        $row[$j] = addslashes((string) $row[$j]);
-                        $row[$j] = str_replace("\n", '\\n', $row[$j]);
-                        $data[] = '"'.$row[$j].'"';
+                        $data[] = $this->quoteForDump($row[$j]);
 
                         if ($j < ($fieldCount - 1)) {
                             $data[] = ',';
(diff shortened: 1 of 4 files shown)


anoziere:
---

Second commit: the `DELIMITER |` rejection is gone too. It existed only because `|` was the placeholder the splitter used internally, so it collided; nothing is reserved any more, and there is a test for a pipe delimiter alongside the one for `//`.

anoziere:
The CI failure on the previous commit was mine, and the cause is worth stating.

`testAFailedRestoreSaysWhichTableItStoppedOn` forced a restore to fail by narrowing a column in the dump and letting its own `INSERT` no longer fit it. That rejection needs `STRICT_TRANS_TABLES`, and Thelia takes that mode off the session on MySQL — which is what CI runs. So the restore succeeded, nothing was thrown, and the assertion had nothing to catch. It passed locally only because the local database was MariaDB, where the same code leaves the mode on (#3779).

Reproduced on MySQL 8.0.40 before touching the test — same test, same `Failed asserting that null is not null` — then confirmed from the cached decision the kernel writes, `'canUpdate' => true` with `Remove sql_mode STRICT_TRANS_TABLES` in its log.

The dump is now broken with a row carrying the wrong number of values instead. Every server refuses that, in any sql_mode and any version:

```
sql_mode=''                    -> ERROR 1136 (21S01) Column count doesn't match value count
sql_mode='STRICT_TRANS_TABLES' -> ERROR 1136 (21S01) Column count doesn't match value count
```

It fails in the most permissive mode there is, so no stricter configuration can swallow it. The test also asserts that the corruption was actually applied to the dump, so it can never again quietly do nothing if the file stops ending the way it expects.

Checked what the coordinator would ask next: the other cases do not rest on the same spring. Under CI conditions — MySQL with the mode removed — **8 of the 19 tests fail on `main`**, and the NULL case fails on the data rather than on an exception:

```
"null in a nullable integer column"   Failed asserting that two arrays are identical
"pipes, as a rich text toolbar holds" Failed asserting that two arrays are identical
"text shaped like an internal marker" Failed asserting that two arrays are identical
testEveryTrapValueSurvivesTheSameDump Failed asserting that two arrays are identical
testANullIsDumpedAsNullRatherThan…    the dump holds "" where NULL was expected
testATruncatedDumpIsRefused…          the truncated dump was replayed: Unknown collation 'utf8mb4_general'
testAFailedRestoreSaysWhichTable…     raw 1136, with nothing said about where it stopped
testAPipeIsAcceptableAsADelimiter     error: the delimiter guard still throws
```

Without the fix the NULL round-trip comes back coerced instead of NULL, so it is red whether or not the server rejects it — permissive mode changes the reason, not the outcome. The truncated-dump case is red in both too, and on MySQL it shows the third state even more plainly: the dump is replayed until it dies on a collation name cut in half.

All five suites are green on MySQL 8.0.40 as well as on MariaDB 10.11: unit 301, integration 622, api 222 (1 skip), flexy 17, backoffice 10. `cs-diff` 0 of 1866, `phpstan` no errors.

PR #3798, opened 2026-08-19
feat: collect the legal identifiers of a business address

NicolasBarbey:
Electronic invoicing requires the legal identifier and the VAT number of a business
buyer to appear on the invoice. Adds a `siret` and a `vat_number` column to `address`,
`cart_address` and `order_address`, next to the `company` name they identify: an
address carries them, a cart address copies them when the cart is built, and an order
freezes them the same way the rest of the address is frozen at checkout.

Both columns are validated by a single shared rule
(`Thelia\Domain\Legal\CompanyIdentifierRules`), consumed by the front and admin
address forms, the customer update form, the Address API resource, and the checkout
guard that keeps an incomplete business address from placing an order. Format rules
are country-aware: a French SIRET/VAT pair is checksummed and cross-checked against
its SIREN, EU member states get their VAT prefix and pattern checked, and everyone
else is stored as free text once it fits the column.

Neither identifier is ever mandatory: an association, a VAT-exempt business or a
company outside the Union legitimately has a name and no registration number to give,
and an address already on file when these columns did not exist yet must stay
editable without retroactively demanding either of them. Only a value that is
actually typed is format-checked.

For existing shops, the new columns ship in `setup/update/sql/3.0.0-beta5.sql`
rather than an already-tagged version: a shop that has already run its update
script never replays it, so it would never pick up the columns otherwise.

Also fixes a LiveComponent regression this feature's testing surfaced:
`Thelia\Core\EventListener\ViewListener` and Symfony UX's `LiveComponentSubscriber`
both listen to `kernel.view` at the same priority, and `ViewListener` is registered
first. It treated every LiveComponent's default render action as an unknown Thelia
view and threw a 404 before `LiveComponentSubscriber` ever got to render it —
breaking every dependent field update and periodic component refresh across the
front office, not just this feature's address forms. `ViewListener` now skips any
request carrying the `_live_component` attribute, the same way it already skipped
LiveComponent's batch-action sub-requests.


anoziere:
Two blockers, plus the PR itself needs some work.

**1. The update script targets beta3, which is already released.**

`setup/update/sql/3.0.0-beta3.sql` ships in the published `thelia/setup` tags 3.0.0-beta3 and 3.0.0-beta4. An install that already ran it stores that version and never replays the file, so no existing shop ever gets the three new columns: only a fresh install does. The ALTERs belong in a new `3.0.0-beta5.sql`. We already had to move statements off `beta2.sql` for exactly this reason in #3711 and #3712.

Related, on the same commit: the `schema.xml` change needs a `thelia/config` extraction and tag before CI can go green on the generated models.

**2. Both identifiers become mandatory on addresses that already exist.**

`CompanyIdentifierRules::violationsFor()` returns a violation as soon as `company` is filled and either field is empty, and that rule is wired into `AddressCreateForm`, `CustomerUpdateForm` and the front and admin API writes. A customer who registered with a company name can no longer update their account without typing a valid SIRET and VAT number, and an integration that has been POSTing addresses for months starts getting 400s. Nothing exempts the existing rows, which contradicts the comment in the SQL script ("Existing rows stay NULL; nothing is required retroactively").

The rule is also stricter than reality: associations, businesses under the VAT exemption threshold and companies outside the EU have a company name and no VAT number. As written, they cannot check out.

**3. No description, and the history needs a pass.**

1600 lines touching the address model, three tables and the checkout, with an empty body. The branch also carries a commit and its own revert (16a6146, 72440fc), worth squashing away. Since we squash-merge, please retitle to the conventional subject that will land on `main`, e.g. `feat: collect the legal identifiers of a business address`.

PR #3831, opened 2026-08-27
fix: answer a write with the translation it saved, not the one it replaced

### LINKED ISSUE #3820: A PATCH on a translatable field answers with the value it replaced
Patching a translatable field saves the new value, but the response body still shows the old one.

## Reproduce

On a demo install, with an admin token:

```
curl -X PATCH https://<host>/api/admin/products/1 \
  -H 'Content-Type: application/merge-patch+json' \
  -d '{"i18ns":{"en_US":{"title":"New title"}}}'
```

The response comes back with the previous title. A GET on the same resource right after returns `New title`, and the back office shows it too, so the write landed.

Patching a field that is not translatable works: `{"ref":"NEW-REF"}` answers with the new ref.

## Expected

The response to a write shows the row as it was saved.

## Where it comes from

The state provider loads the model with the translations joined in, so `product_i18n.title` sits in a virtual column of that model instance. The write sets the new value and saves it, but the virtual column filled by the earlier query still holds the old string, and `ApiResourcePropelTransformerService::manageTranslatableResource()` prefers the virtual column over the getter:

```php
if ($baseModel->hasVirtualColumn($virtualColumn)) {
    $fieldValue = $baseModel->getVirtualColumn($virtualColumn);
} else {
    $propelModel->setlocale($lang->getLocale());
    $getter = 'get'.ucfirst($i18nFieldName);

    $fieldValue = $propelModel->{$getter}();
}
```

Trusting the joined column is right for a read: an empty column means the language has no translation, and falling back to the getter costs a query per row and per language. On an instance that a write has just changed, the column is out of date, while the getter returns the new value.

Reproduced on `main` and on 3.0.0-beta5 with the demo data.

anoziere:
Fixes #3820

## What happened

A `PATCH` on a translatable field saved the new value but answered with the one it replaced.

```
PATCH /api/admin/categories/12
{"i18ns":{"en_US":{"title":"Title after the patch"}}}
```

came back with `Title before the patch`, while the row already held the new one.

**The write was never at fault.** The database, a later `GET` and the back office all showed the new value. Only the response body was behind, which is why it could go unnoticed for a while: a client that trusts the answer to its own write ends up displaying stale content until it reloads.

## Why

The item provider reads the model with every active language left-joined, so `category_i18n.title` sits in a virtual column of that instance. Propel's instance pool then hands the *same* object to the write, which sets the new title and saves it — the virtual column filled by the earlier query is never refreshed and keeps the old string.

Building the response, `ApiResourcePropelTransformerService::manageTranslatableResource()` prefers that column over the translation getter. On a read that preference is right and cheap: the join answered for every language, so an empty column means "no translation", and falling back to the getter costs a query per row and per language. On the instance a write has just changed, the column is simply out of date.

## The fix

The write response now tells the transformer that the joined i18n columns belong to the query that ran *before* the write, so it reads the translation getters instead — the only source that knows what was just saved.

Reads are untouched: the eager-loaded columns are still preferred everywhere else, so the N+1 work done in #3812 / #3813 / #3814 stands. The extra queries are paid once, on the response to a single write.

## Test

`CategoryApiTest::testPatchCategoryAnswersWithTheSavedTranslation` patches one locale of a two-locale category and asserts the response carries the new title, that the untouched locale is still there, and that the row matches once the instance pool is cleared.

Red before the change:

```
Failed asserting that two strings are identical.
-'Title after the patch'
+'Title before the patch'
```

Green after.

`composer cs-diff`, `composer phpstan` and the `unit` / `api` / `http-flexy` / `http-backoffice` suites pass.

### DIFF
--- a/core/lib/Thelia/Api/Bridge/Propel/Service/ApiResourcePropelTransformerService.php
+++ b/core/lib/Thelia/Api/Bridge/Propel/Service/ApiResourcePropelTransformerService.php
@@ -39,6 +39,13 @@ use Thelia\Model\Lang;
 
 readonly class ApiResourcePropelTransformerService
 {
+    /**
+     * Context flag telling the transformer that the model instance it reads
+     * carries i18n columns joined before a write, so they hold the values that
+     * write replaced. Set it on a write response, never on a read.
+     */
+    public const STALE_I18N_VIRTUAL_COLUMNS = 'thelia_stale_i18n_virtual_columns';
+
     public function __construct(
         private array $apiResourceAddons,
         private ValidatorInterface $validator,
@@ -676,6 +683,8 @@ readonly class ApiResourcePropelTransformerService
         array $context,
         Collection $langs,
     ): void {
+        $staleVirtualColumns = (bool) ($context[self::STALE_I18N_VIRTUAL_COLUMNS] ?? false);
+
         foreach ($langs as $lang) {
             $i18nResource = new ($resourceClass::getI18nResourceClass());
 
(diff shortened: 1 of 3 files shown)


anoziere:
The red `test` check does not come from this branch: `BaseFormCsrfTokenTest::testTheDefaultTokenIsBoundToTheSessionAndRoundTrips` already fails on `main` at cc09ed8 (run 33069882090), and it fails identically here with both changed files reset to `HEAD`. The `unit`, `api`, `http-flexy` and `http-backoffice` suites are green, as are `cs-diff` and `phpstan`.

PR #3913, opened 2026-09-08
fix(routing): put the rewriting router listener strictly first

anoziere:
## Summary

- Two `RouterListener` objects answer `kernel.request` at priority 32: the one the core registers around the chain router, which is the only one that can consult `router.rewrite`, and the one FrameworkBundle registers around `router`.
- Whichever runs second returns immediately, `_controller` being set already, so the one that runs first is the one that routes. At equal priority that came down to the order the two services were registered in — nothing guaranteed the rewriting router would ever be consulted.
- The core listener now declares its three events one by one, with `kernel.request` one point in front of FrameworkBundle's and the other two at the priorities of the class.
- No measurable cost either way: this is about which listener decides, not about how long it takes.

FrameworkBundle's listener is deliberately left in place: it is the one that keeps `router.request_context` in step with the request, and `router.default` generates URLs from that context.

## Test plan

- [x] `tests/Integration/Core/Routing/RouterListenerPriorityTest.php` — the first router listener to run is the one holding the chain router (an invariant that used to hold only by luck), and the events declared in the config match what `RouterListener::getSubscribedEvents()` asks for, with `kernel.request` strictly in front. The second assertion is red before the change and green after.
- [x] `debug:event-dispatcher kernel.request` shows the core listener at 33 and FrameworkBundle's at 32.
- [x] `composer test` green (unit 439, integration 892, api 341, http-flexy 89, http-backoffice 17)
- [x] `composer cs-diff` clean
- [x] `composer phpstan` no errors

### DIFF
--- a/core/lib/Thelia/Config/Resources/services/core/routing.php
+++ b/core/lib/Thelia/Config/Resources/services/core/routing.php
@@ -16,6 +16,7 @@ namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\HttpKernel\EventListener\RouterListener;
+use Symfony\Component\HttpKernel\KernelEvents;
 use Thelia\Controller\Admin\BaseAdminController;
 use Thelia\Core\Routing\Loader\XmlFileLoader;
 use Thelia\Core\Routing\ModuleAttributeLoader;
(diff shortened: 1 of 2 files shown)


anoziere:
Rebased onto main (the branch had drifted 5 commits behind) and reran everything locally: `composer test` green across all five suites, `cs-diff` and `phpstan` clean, and the two guest-checkout theme packages updated to the versions CI resolves (1.0.3). All 17 http-backoffice tests pass locally under that setup, including against a real MySQL 8 instance (not just the local MariaDB), matching CI's engine as closely as I could reproduce it outside GitHub Actions.

CI itself stays red on this branch specifically: `ConfigStoreGuestCheckoutTest` (2), `CustomerGuestCheckoutTest` (2) and `ProductGuestCheckoutForbiddenTest` (1) get a 500 instead of a 200/302, on PHP 8.3, 8.4 and 8.5, both before and after the rebase. The same five tests pass on main's own CI. Neither test surfaces the underlying exception (no `catchExceptions(false)`), and I could not get a stack trace out of the failed-job log.

Given the change here is only reordering two `kernel.event_listener` priorities so the core's rewriting router chain wins over FrameworkBundle's plain one, and that this makes `RewritingRouter` (previously reached by chance, per the PR description) consulted on every request where it used not to be, I suspect an interaction between that and one of these three back-office screens under the current theme — but I have not been able to pin it down within a reasonable amount of investigation, and would rather not push a fix I cannot prove correct.

Closing per the standing instruction to close rather than guess when a fix cannot be made safely. Findings above are for whoever picks this back up: reproduce with `catchExceptions(false)` added to the three failing tests to get the real exception first.

PR #3920, opened 2026-09-09
feat(catalog): give the relation between products a configurable type

NicolasBarbey:
## What

The relation between two products now carries a configurable type. Until now Thelia knew a single
one — the accessory — and nothing told an accessory apart from a complementary product or a
higher-end model.

Three types ship and are active on a fresh install: `accessory`, `cross_selling` (reciprocal) and
`up_selling`, each with a translatable label and a stable code. The table `accessory` is kept as it
is, columns and all, and gains `type_id`: every relation already saved takes the accessory type, and
modules that query `AccessoryQuery` directly keep working.

- `product_association_type` (+ i18n): `code` unique, `visible`, `reciprocal`, `position`.
- `accessory.type_id`, NOT NULL, foreign key in `RESTRICT` — deleting a type a shop still uses is
  refused rather than silently taking its relations along.
- Index `(product_id, type_id, position)`, to serve one block of a product sheet in a single read.
- `Accessory::addCriteriaToPositionQuery()` now scopes the position to the product **and** the type,
  so reordering one block leaves the others where they were.
- Facade: `addAssociation()`, `removeAssociation()`, `getAssociations(int, ?string)`. `addAccessory()`
  and `removeAccessory()` are kept and delegate with the accessory type.
- Events `PRODUCT_ADD_ASSOCIATION`, `PRODUCT_REMOVE_ASSOCIATION`,
  `PRODUCT_UPDATE_ASSOCIATION_POSITION`. The three accessory events are kept and still fire.
- Self-relation refused, in the facade and in the listener.

## `accessory` loses `isCrossRef` — the one API break

Propel supports exactly two foreign keys on a table declared `isCrossRef="true"`. With a third it
generates the `Product ↔ ProductAssociationType` many-to-many twice and the model stops compiling
(`Cannot redeclare Thelia\Model\Base\ProductQuery::filterByProductAssociationType()`). A junction
table that carries data of its own is no longer a pure junction, so the flag goes.

What this drops: 24 generated many-to-many methods on `Product` / `ProductQuery`
(`getProductsRelatedByAccessory()`, `addProductRelatedByAccessory()`,
`filterByProductRelatedByProductId()`, …). None of them is called anywhere in `core/`, `src/`,
`tests/`, the three bundled themes, `local/modules/` or `vendor/thelia` — checked before deciding.
`Accessory` and `AccessoryQuery` are untouched. A third-party module calling one of those generated
methods would break, hence flagging it here.

## Reciprocity is written by the listener, and announced exactly once

Adding a relation of a reciprocal type writes the mirror row and announces it. Two flags on the
event keep the announcements honest:

- `applyReciprocity` stops the mirror from mirroring back.
- `announceAccessoryEvent` stops the legacy accessory event and the new association event from
  re-entering each other.

Without the second one, a single add would announce `PRODUCT_ADD_ASSOCIATION` twice on the same
direction — the write would stay correct thanks to the existence check, but a module counting
additions would count the same relation twice. Each direction is announced once, in both
directions, and two integration tests pin it.

## Migration

`setup/update/sql/3.0.1.sql`, in three steps: create the type table and seed the three types, add
`accessory.type_id` **nullable** and backfill it with the accessory type, then require the column and
add the key and the index. Adding a NOT NULL column with no default to a table that already has rows
is refused outright in strict mode, which is what any shop with accessories would have hit.

Replayed on a database rebuilt in its pre-migration state: the existing relations take the accessory
type and keep their positions, and a second run duplicates no type, no i18n row, no constraint and no
index.

## Managing the types, and the screen behind them

The second commit hands the types themselves to the merchant: five events —
`PRODUCT_ASSOCIATION_TYPE_CREATE`, `_UPDATE`, `_DELETE`, `_UPDATE_POSITION` and
`_TOGGLE_VISIBLE` — and the action that answers them, modelled on the checkout consents, which
had the same shape of screen.

The **code** of a type is set once and never changes. The core names the accessory type by its
code so `addAccessory()` keeps meaning what it always meant, and the front-office blocks are keyed
on it; `update()` ignores the field rather than letting a rename break both. The translatable title
is what a merchant rewords instead.

Deletion is refused twice over: for the codes in `ProductAssociationType::UNDELETABLE_CODES`, and
for a type still carrying relations. The foreign key would refuse the second one anyway, with a
Propel error nobody can act on. Hiding a type is the answer to "stop offering this", and it keeps
the relations already saved under it.

The back-office screen that drives all this — the relations tab split per type and the type
configuration screen — lives in the `default-twig` theme, which is its own package:
**thelia-templates/default-twig#74**. It has to be merged after this one.

## The API resource

The third commit publishes the relations over the API, front and admin:
`ProductAssociation` and `ProductAssociationType`, with its i18n resource.

- Writes go through the facade, and the types through the five events above, so the
  three guards — the code set once, the undeletable codes, a type still carrying
  relations — hold over the API exactly as they do in the back-office.
- `ProductAssociationVisibilityExtension` bounds the front to visible products
  **and** visible types, server side. That is stricter than `/front/products`,
  which leaves it to the caller to ask for `visible=true`; a block on a product
  sheet has no reason to offer either.
- Two entries in `AdminApiResourcePermissions`. The admin API refuses by default, so
  a resource missing from that list is refused to everyone, superadmin included.
- `position` is read only: reordering already has its event and its back-office
  screen.

### One read per block

A block was one query per relation. `EagerLoadingExtension` joins the many-to-one
end but only adds its columns, and nothing reads them back, so the transformer asked
the Propel getter and paid a `findPk` a row — product selects went 2, 4 then 9 for
1, 3 and 8 relations. The Propel instance pool hides this for as long as the rows
point at the same target, which is why nothing showed.

`PropelRelationPreloader` already reads such an end for a whole page, and already
tells apart two foreign keys towards the same table; a guard held it back on the
belief the query had hydrated them. The `Relation` attribute gains `preload`, and
`ProductAssociation` is the only resource that sets it, so no other resource changes
behaviour. A block of two relations and then of six reads in 15 queries against 27
without it.

The related product comes out whole rather than as a bare IRI — the wording, the
rewritten url, the sale elements with the default one told apart by `isDefault`, and
its price — which is what a card of a block shows.

This is not the dedicated provider the task asks for, and the reason is worth
recording. Propel refuses `with()` on a one-to-many combined with `limit()`, so on
any paginated collection: the join on the default sale element cannot be written
into the query the way the task phrases it. And a `with()` collides with the virtual
columns `EagerLoadingExtension` already sets, on `Error populating Product object`.

### A correlation bug fixed on the way

Reaching the same translatable resource through two relations of one parent served
the second one the translations of the first: a related product came back under the
title of the source product. `EagerLoadingExtension::joinI18ns()` names its virtual
columns after the **property** the relation is read through, while the transformer
read them back by the **resource** name, and the two only agree when a property
bears the name of its target. The fix touches shared infrastructure, and the five
suites cover it.

## Out of scope

The front-office blocks are a separate task, and they live in the Flexy theme's own
package. Two known consequences remain:

- `getAssociations()` does not filter on the type's `visible` flag: the signature the
  task specifies has no room for it. The API bounds the front to visible types
  instead, which is the path a shopper-facing read actually goes through.
- The i18n filter of the API has no language fallback, so a type a merchant created
  in a locale they left empty comes back with no title. That is upstream of this
  branch and older than it.

Also worth stating: relating a product to itself **was** possible on the accessory path, which had no
guard at all. Refusing it is a deliberate behaviour change.

## Dependency

The column and the table need a `thelia/config` release. Until it ships, a `composer update`
overwrites `local/config/schema.xml` and the column disappears from this branch.

## Tests

53 new tests. 16 integration tests on the relations themselves
(`tests/Integration/Action/ProductAssociationActionTest.php`): add per type, positions counted per
type, reordering one block only, mirror row, one announcement per direction, removal of both
directions, self-relation refused, unknown type refused, cascade on product deletion, the accessory
path still writing under its type, and the facade end to end. Then 8 integration tests on the type
action (`tests/Integration/Action/ProductAssociationTypeActionTest.php`) and 7 HTTP tests on the
back-office screen (`tests/Http/BackOffice/ProductRelationTypesTest.php`): the CRUD, the refusals,
and the code left alone by an update.

Then 19 API tests, 11 admin and 8 front
(`tests/Api/{Admin,Front}/ProductAssociationApiTest.php`): the seven return codes,
the bounds the front is held to, the embedded product, and the query counts of a
block read twice. The visibility extension was switched off to check that the three
tests pinning it do fail without it. And 3 HTTP tests on the front blocks
(`tests/Http/Flexy/ProductRelationBlocksTest.php`): one block per filled type in the
order the types are arranged, no title left behind by a type nothing was picked
under, and a product taken offline offered by no block.

All five suites pass, `composer cs-diff` is clean, and both PHPStan configurations report no errors —
run on each commit as it was written, the last time against `main` at `1bf93d2`. The state this branch
is in now, merged with `main` and squashed, is what CI is being asked to confirm.
The fresh install and the demo database were checked: the three types are present with their French
and English labels, and the demo accessories are related under the right type.

Rebased on `main` after #3921, then merged with `main` again to take #3927. The two branches both append to `setup/update/sql/3.0.1.sql`, and
each `SET @statement` there keeps its own `PREPARE` / `EXECUTE` / `DEALLOCATE`: concatenating the
two blocks would have overwritten the variable and silently dropped an `ALTER TABLE`.

## What CI covers here, and what it skips

The `SaleCustomerIdsReader` failures this branch used to inherit from `main` are
gone: #3927 resolves the themes from `dev-main` rather than from their releases, so
the core is now proven against the theme branch tips.

That has one consequence worth stating plainly. The two front-office suites guard on
the product page reading `/api/front/product_association_types`, and the blocks that
read it live on the theme branch of the front-office task, which is not merged. So
`ProductRelationBlocksTest` reports its 3 tests as skipped — and so does
`ProductAccessoriesTest` with its 4, because the strip it used to look for by its
English title is one of these blocks now and its guard follows the same read. On the
theme branch tip those 4 would otherwise have run. Both suites come back on their own
the moment the front-office side merges; locally, with that branch mounted, all seven
pass.


anoziere:
I merged `main` into the branch and pushed ten commits on top of it. Nothing in your work was reverted; each commit fixes one point found while reviewing the branch against the acceptance criteria of the issue, with a test that failed before the fix and passes after it.

**Security (front API)**

- `478a4ff19` The front relation endpoints exposed products of a reserved sale (private drop) to anonymous visitors. The visibility rule is now applied on both ends of the relation.
- `20bd6374d` Same endpoints served the relations of an offline product, with the full product sheet (i18n, price, public URL) and a 200 on the single read instead of a 404. The `visible` rule was only checked on the accessory side.

**Admin API validation**

- `10baef2b2` A relation posted without its type or without one of its two products was accepted; it is now refused.
- `9cf5a1553` A duplicate type code failed on the unique index (500); it is now refused with a 422.
- `ac104bbb5` A type created without any title was accepted; it now needs a title.
- `a21da1986` The title check only looked at the first locale of the payload, so a type titled in French but described in English first was refused. Every posted locale is now considered.

**i18n / PATCH**

- `94e038955` Rewording a type reset its `visible` and `reciprocal` flags.
- `01d1e0845` A flags-only PATCH erased the wording in every language.
- `ece316980` A title-only PATCH set the description to NULL.

`1e6ede358` moves the regression tests into the existing test classes (`tests/Api/Admin/ProductAssociationApiTest.php`, `tests/Api/Front/ProductAssociationApiTest.php`, `tests/Integration/Action/ProductAssociationTypeActionTest.php`).

Against `main`: code style and PHPStan unchanged, +10 api tests and +2 integration tests, no new failure, no new skip. The ten acceptance criteria of the issue that live in the core are covered by tests or by the install and update scripts (three types seeded in fr/en, idempotent update, positions per type, accessory path and events preserved, reciprocity, self-relation refused, no orphan, bounded front read, admin API through the facade, constant query count per block). CI is green on `a21da19`; the PHPStan job of the back-office theme went red on PHPStan 2.2.14 released the day before, fixed on the theme's `main`.

Four points the issue left open, and how we close them:

1. Reciprocity is a setting of the type, as implemented, and not a checkbox per relation. Kept as is.
2. `accessory` loses `isCrossRef`, which Propel imposes once a table has a third foreign key. The 24 generated many-to-many methods on `Product` and `ProductQuery` (`getProductsRelatedByAccessory()`, `addProductRelatedByAccessory()`, ...) disappear. Nothing in the core, the themes or the published modules calls them, but a third party module might. Listed as a breaking change in the 3.1 notes rather than restored by hand.
3. The back-office screens (thelia-templates/default-twig#74) and the Flexy blocks (thelia-templates/flexy#203) are reviewed together with this branch, on the same install. #74 currently conflicts with `main` of its theme on `translations/messages.fr_FR.php` and needs a rebase first.
4. The new table and the `accessory.type_id` column need a `thelia/config` release before publication. Merge first, release grouped with the other catalog columns of the lot, as the issue suggests. Until then a `composer update` on a dev install rewrites `local/config/schema.xml` and drops the column.

PR #3940, opened 2026-09-11
feat(customer): native customer tags

pierrebaguet:
Adds customer tags to the core: free-text markers an administrator puts on a customer, used to filter the back-office customer list and, later, to target promotions.

The back-office screens are in a companion PR on `thelia-templates/default-twig`. **Merge this one first**: the theme calls `TagService` and catches `TagLabelAlreadyUsedException`. The HTTP tests added here skip themselves when the installed theme has no tag screens, so this branch is green on its own.

## What it adds

Two tables. `tag` holds the vocabulary (label, optional `#RRGGBB` colour). `tag_element` attaches a tag to an object by an element key and an identifier, so the same tag reaches customers today and orders or products later without a second pivot table.

`Thelia\Domain\Tagging\Service\TagService` is where the screens go through: `create`, `rename`, `attach`, `detach`, `setLabelsFor`, `merge`, `findTagsFor`, `findTagsForMany`, `countCustomersByTag`, `countCustomersCarryingEither`, `pruneOrphanedCustomerLinks`.

Two API resources under `/admin/tags` and `/admin/tag-elements`, with no front operation. `Tag` is guarded by its own `AdminResources::TAG`, `TagElement` by `AdminResources::CUSTOMER`: a profile allowed to rename or delete a tag reaches every customer carrying it, which is a wider reach than editing one customer.

A `tag:prune-orphans` command, plus cleanup in `Customer::postDelete()` and in `CustomerAnonymizer`.

## Decisions worth a second opinion

**Uniqueness sits on `label` alone, with no normalised column.** Measured on MariaDB 10.11: under `utf8mb4_general_ci`, `'VIP' = 'vip'` and `'Salon' = 'Salón'` are both true, while `'VIP  club' = 'VIP club'` is false. The collation already folds case and Latin accents, so a normalised column in the same collation would reject nothing more. `Tag::preSave()` collapses inner whitespace, which the collation does not. An integration test inserts the accented pair and expects the unique violation, so the rule is pinned rather than inherited from a database setting.

**`tag_element` has no foreign key on `element_id`**, because the target is polymorphic. `meta_data` in core has the same shape and leaks: nothing anywhere deletes its rows when the object goes. Hence `Customer::postDelete()` for every deletion that goes through the object, and `tag:prune-orphans` for the bulk deletes and hand-written SQL that never reach it.

**Anonymisation does not use `CustomerPersonalDataProviderInterface`.** `CustomerPersonalDataExporter::export()` writes a section for every provider, including one that returns an empty array, so a `"tags": []` section would appear in the archive handed to the customer. Tags are core rather than a module, so the cleanup goes straight into `CustomerAnonymizer`, in the same transaction as the rest.

**The migration goes in `setup/update/sql/3.1.0.sql`**, the pending update script, rather than in a new file. It already accumulates several unrelated features, and it was renamed from `3.0.1.sql` while this branch was open, so the branch follows. Every statement in it is idempotent, and the resource row is seeded by its code with `INSERT IGNORE` rather than by a fixed id.

**`element_key` is not restricted to a list of allowed values.** A module has to be able to declare its own key, and it cannot extend a PHP attribute written in core. The residual risk is an attachment pointing at a key nobody reads, in a resource already restricted to administrators.

## Tests

101 new tests: 35 on the domain service, 17 on the two API resources, 47 on the back-office screens (these skip without the companion theme), and 2 on anonymisation and the personal data export. Every one of them was seen red against deliberately broken code before being kept.

The five suites, cs-diff and phpstan are green on PHP 8.3, 8.4 and 8.5. The 47 back-office tests skip in CI until the companion theme is merged, because the job installs the themes from their `main` branch; they were run green locally against the theme branch.

## Not covered

The back-office screens ship French translations only. The JavaScript tag picker in the companion PR is verified in a browser, not in CI.


anoziere:
Pushed 4 commit(s) on this branch:
- An existing test is red on the head of the branch: Thelia\Tests\Unit\Install\UpdateScriptTest::testThePendingUpdateScrip (2f3963d80)
- POST /api/admin/tags and PATCH /api/admin/tags/{id} answer HTTP 500 when the label is one another tag already carries: t (c015c5037)
- An admin profile holding only AdminResources::CUSTOMER writes the shop tag vocabulary: posting thelia_customer_update[ne (templates/)
- The collision guard added in c015c5037 refuses PUT /api/admin/tags/{id} with 422 whenever the body carries the label the (c6f608778)

PR #3942, opened 2026-09-14
Configurable checkout steps and display mode

mabruchet:
The checkout tunnel is now described by configuration instead of theme code. A `checkout_step` table (plus i18n) carries the order, activation and wording of the steps; what a step does lives on `CheckoutStepProviderInterface` services tagged `thelia.checkout.step_provider`, so a module can ship its own step. `CheckoutProgressionService` is the single authority on where a cart stands (`activeSteps`, `firstIncompleteStep`, `isReachable`) — session-free, memoised per cart state. A `checkout_display_mode` setting publishes the display form (steps or one page); rendering stays in the theme.

Turning a step off removes its screen, never its check: placement runs the `check()` of every registered step provider — a module's step is enforced at `/checkout/pay`, not only on screen. A broken configuration falls back to the defaults with a log warning. Seeds keep the historical wordings, and a cart with nothing to ship no longer sees the delivery screen, as the story asks.

Fresh installs get the four current steps; the migration is idempotent (`3.1.0.sql` — deploy the SQL before the code, a missing table falls back to the providers' defaults). Needs thelia/config#13 merged and tagged first. Consumed by thelia-templates/default-twig#82 (configuration screen) and thelia-templates/flexy#207 (rendering, both forms). Developer map in `docs/checkout-steps.md`.


anoziere:
Pushed 2 commit(s) on this branch:
- Le repli "tunnel invendable" de orderedCodes() rejoue les memes defaultPosition() que celles qui ont casse la forme : il (8dd065039)
- Tunnel une page : getSections() ne demande que mayEnterCheckout(). Un invite deja identifie dont le panier gagne un prod (ae4a00714 )
Gate deltas are in the description.

anoziere:
Correction to my note above: no quality gate ran with those two commits. The review stopped before that phase, so nothing in the description carries gate deltas, and the same goes for the note on thelia-templates/flexy#207.

I ran the five suites by hand on the same install afterwards, and pushed a third commit (`e6c583b3b`) for a defect the review had reproduced red and left open. A step a module declares but the shop has not synchronised yet was served in the right place — between the cart and the payment — while `CheckoutStepView::position` still carried its raw `defaultPosition()`. Rank and position contradicted each other (cart 1, delivery 2, the module step 10, payment 3), so a theme that numbers or sorts the trail on `position` put that step behind the confirmation. It now carries the rank it is served at, with a regression test in `CheckoutProgressionTest`.

Where the branch stands, measured on an install with the two theme branches mounted: 2497 tests, 1 skipped, 0 failures (unit 590, integration 1236, api 440, http-flexy 162, http-backoffice 69), php-cs-fixer clean, PHPStan clean. The upgrade path was replayed on a 3.0.0-shaped database: `3.1.0.sql` creates the two tables, seeds the four steps with their wordings, the config key and the admin resource, and a second pass changes nothing and leaves a merchant's own order and activation alone.

`thelia/config#13` is merged and tagged `3.1.0`.

PR #3956, opened 2026-09-16
build(propel): require thelia/propel ^1.0.3

anoziere:
## Why

On PHP 8.5 the test suite raises 546 deprecation notices, against 0 on PHP 8.3 and 69 on PHP 8.4 (measured job by job on the same commit). All 546 come from the Propel generator emitting non-canonical cast names, deprecated in 8.5:

- `ObjectBuilder` wrote `(boolean)` in the boolean mutator, and `(` . `Column::getPhpType()` . `)` in the primitive mutator, in `hydrate()` and in the lazy-load hydrator — `getPhpType()` returns the PHPDoc spelling (`boolean`, `double`), not the cast keyword
- `TableMapBuilder` did the same for the two primary key casts
- `ModelJoin` called `ReflectionProperty::setAccessible()` and `SchemaReader` called `xml_parser_free()`, both no-ops since 8.1 and 8.0 respectively, both deprecated in 8.5

`thelia/propel` 1.0.3 routes the casts through the `PropelTypes::getNativeTypeDeclaration()` table already present in the package and drops the two dead calls. `(bool)` and `(float)` are strict aliases of `(boolean)` and `(double)`, so the generated models behave identically.

## Result

Models regenerated against 1.0.3 contain no non-canonical cast at all. Rerunning the suites on PHP 8.5:

| Suite | before | after |
|---|---|---|
| unit | 17 | 1 |
| integration | 186 | 55 |

What remains is a different family — `Implicitly marking parameter … as nullable`, deprecated in **8.4**, already present today and raised by third-party modules (`Page`, `Tinymce`, `CustomerFamily`, `RewriteUrl`, `GoogleShoppingXml`) and by `thelia/currency-converter` and `commerceguys/addressing`. None of it comes from the core; it is out of scope here.

## Note

Raising the floor to `^1.0.3` rather than leaving `^1.0` is what forces an existing lock off 1.0.2.


anoziere:
Carried by the 3.1.0 release branch

PR #3974, opened 2026-09-18
Add choice filter on brand page

robinallezard:
A brand page previously had no filters: Thelia organizes filters by category, and a list without a category displays none. It now uses the columns from the categories where its visible products are organized, broken down by what each row governs (a feature, an attribute, or another criterion). When two categories govern the same criterion, the row selected is the visible one, in the lowest position. The values displayed are those associated with the brand’s products, along with their quantities; thus, there are no filters that lead nowhere.

The “brand” facet is never offered on a brand page: the list is already that of a specific brand, and selecting another brand would return no products.

## How a Page Identifies Itself

getFilters() reads a scope parameter alongside tfilters: the page for brand 3 sends scope[brand]=3.

This declaration is necessary because `category=7 & brand=3` represents the same selection for a category page where a brand has been selected and for a brand page where a category has been selected, even though the two require opposite columns. A caller that does not declare a `scope` retains the previous interpretation: a single brand, outside of a category and outside the list of identifiers, remains a brand page.

## Cost

The `choice_filter` lines for all categories are read in bulk, with one query per level of the tree to trace back to the parent category templates. The number of queries no longer follows the brand’s distribution. 

FLEXY PR => https://github.com/thelia-templates/flexy/pull/220


anoziere:
Pushed 2 commit(s) on this branch:
- nearestCategoriesHavingTemplate() keeps only the single nearest template-carrying ancestor per category, so choiceFilter (285ec08c1)
- scope[brand] lifts the guard of getFilters() without ever narrowing the query: ?scope[brand]=1 and ?scope[brand]=2 on /a (56bddf359)

PR #3977, opened 2026-09-18
Catalog price rules by scope and period

openstudio-sdiakite:
A merchant writes "20% off the Winter category in January" or "9.90 on brand X for these customers" and the shop prices follow, in the listings, on the product page, in the cart, on the order and through the front API. Nothing is written into `product_price`: the rule keeps the prices a merchant typed and the ones a flash sale wrote untouched, and stops the moment it ends.

A rule has a scope (categories with or without their descendants, brands, templates, feature values, attribute values, named products, combined with AND), an audience (everyone or named customers; customer groups are modeled for a later story) and an effect (percentage, amount per currency or fixed price per currency, tax included at the shop location like the sale offsets), plus a priority, a stop flag and dates.

The scope is materialized per rule, and the public price is stored as dated validity segments, so an opening or a closing takes effect at the second without a cron. Named customers are priced at read time with the same engine. Catalog changes, back-office and API, re-evaluate the product they touch; `catalog-price-rule:recompute` catches up what a change too large for its request left over. A flash sale converts into turned-off rules.

Decision taken with the product owner: the rule replaces the displayed promo price, a manual promo or a running flash sale included.

Breaking changes are listed in the changelog (loop, cart and access service constructors, `ReservedSalePriceListener` renamed, cache parameter renamed with an alias). The back-office screens are in the companion pull request thelia-templates/default-twig#92. `docs/catalog-price-rules.md` is the map.

The update script is `setup/update/sql/3.2.0.sql`; `TheliaKernel::THELIA_VERSION` is left for the release to bump.


anoziere:
Reviewed this branch against the linked issue. No change pushed.
3 question(s) still need a decision before this can be merged.

PR #3978, opened 2026-09-18
Checkout over the front API: selections, validation and idempotent order placement

mabruchet:
A headless front can now buy end to end through the front API. Under `/api/front/account/checkout/{cartId}`, four POST operations set the delivery address, invoice address, delivery module and payment module — each returns the recalculated cart — `GET …/validation` lists every blocking point at once with stable machine codes, and `POST …/place` writes the order, decrements stock, sends the confirmation emails and tells the caller what the payment expects: nothing, a redirect URL, or a gateway form. Consents are answered in the placement body and frozen onto the order exactly as the theme does.

The API walks the same checkout rules as the theme: `validateForOrder()` and the step providers are untouched, violation collection is additive. Placement is idempotent — a cart that already carries a live order returns it (`alreadyPlaced: true`) instead of creating a second one, guarded by a lock and a transactional re-read on `order.cart_id`. Every operation requires an authenticated customer who owns the cart; an unknown cart and a foreign cart get byte-identical 404s, and amounts, shipping and discounts are never read from the caller.

Also fixed on the way: coupon listeners and `AbstractPaymentModule::generateGatewayFormResponse()` no longer crash on a session-less request, a stock shortage is now a typed `StockShortageException` naming the product reference, and the checkout tunnel messages are translated in every shipped locale. Behaviour notes are in the changelog, including that an `Order` carrying a currency or lang now wins over the session in `ORDER_PAY`.


anoziere:
Pushed 1 commit(s) on this branch:
- The cart answered by the four checkout operations is taxed in the wrong country: normalize() takes it from TaxEngine::ge (9c618f292)
2 item(s) could not be fixed automatically and are listed there too.
4 question(s) still need a decision before this can be merged.

